PR #1, opened 2020-07-10
CCA logic

kdamball:


lib/create-app-definition.js
```diff
@@ -0,0 +1,91 @@
...
+    console.log();
+    console.log(`App ${appName} has been successfully created in ${selectedOrg.name}`);
+    console.log();
+    console.log();
+    console.log('\033[1mNext steps:\033[0m ');
+    console.log(
+      `   1. To develop, run \`npm start\` inside your app folder and open https://app.contentful.com/deeplink?link=apps&id=${createdAppDefition.sys.id}`
+    );
+    console.log(
+      `   2. To learn how to build your first Contentful app, visit https://ctfl.io/app-tutorial`
```
Jwhiles, line 79: This link doesn't work. Though I assume you know this :)

template/src/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  const root = document.getElementById('root');
+
+  // All possible locations for your app
+  // Feel free to remove unused locations
+  // Dont forget to delete the file too :)
+  const ComponentLocationSettings = [
+    { location: locations.LOCATION_APP_CONFIG, component: <Config sdk={sdk as AppExtensionSDK} /> },
+    {
+      location: locations.LOCATION_ENTRY_FIELD,
+      component: <Field />
```
Jwhiles, line 26: I'd still like to the correct sdks passed to these components, even if they aren't used. But understand if you don't want to 🤷

lib/create-app-definition.js
kdamball: Nope - thanks for catching it.

template/src/index.tsx
kdamball: incoming changes :)

.eslintignore
```diff
@@ -0,0 +1,3 @@
+**/node_modules/**
+/template/src**
+/template/bin**
```
jelz, line 3: What's that?

lib/create-app-definition.js
```diff
@@ -0,0 +1,91 @@
...
+
+  const { organizationId } = await inquirer.prompt([
+    {
+      name: 'organizationId',
+      message: 'Select an organization for your app:',
+      type: 'list',
+      choices: organizations
+    }
+  ]);
+  selectedOrg = organizations.find(org => org.value === organizationId);
```
jelz, line 44: you could do `const selectedOrg` here

lib/create-app-definition.js
```diff
@@ -0,0 +1,91 @@
...
+      `   1. To develop, run \`npm start\` inside your app folder and open https://app.contentful.com/deeplink?link=apps&id=${createdAppDefition.sys.id}`
+    );
+    console.log(
+      `   2. To learn how to build your first Contentful app, visit https://ctfl.io/app-tutorial`
+    );
+    console.log();
+  } catch (err) {
+    console.log();
+    console.log(
+      'Something went wrong with creating app definition. Run `create-contentful-app create-definition` to try again.'
```
jelz, line 85: "when creating an app definition"

lib/index.js
```diff
@@ -0,0 +1,110 @@
...
+const command = process.argv[2];
+const appFolder = process.argv[3];
+
+const mainCommand = 'npx create-contentful-app';
+
+function onSuccess(spinner, folder) {
+  spinner.stop(true);
+
+  console.log();
+  console.log(`Success! Created "${folder}" in ~/${folder}.`);
```
jelz, line 20: Is "in ~/${folder}" part true? I'd just say "Created (full path here)"

lib/index.js
```diff
@@ -0,0 +1,110 @@
...
+    console.log('Available commands:');
+    console.log();
+    console.log(`${mainCommand} init app-name`);
+    console.log(`   Bootstraps your app inside of a new folder "app-name"`);
+    console.log();
+    console.log(`${mainCommand} create-definition`);
+    console.log(`   Creates an app definition for your app.`);
+    console.log(`   This will prompt you to login into Contentful,`);
+    console.log(
+      `   Paste your access token after obtaining it and select which organization app definition should be installed in`
```
jelz, line 85: - paste - lowercase,
- "select in which organization the app definition"
- installed -> created. Apps are installed in environments, definitions are created

lib/index.js
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env node
+
+/* eslint-disable no-console, no-process-exit */
+
+const spawn = require('cross-spawn');
+const path = require('path');
+const Spinner = require('cli-spinner').Spinner;
+const fetchAppDefinitionSettings = require('./prompt-app-definition');
```
jelz, line 8: Is it really "fetch"? "prompt" sounds better

lib/prompt-app-definition.js
```diff
@@ -0,0 +1,59 @@
+const inquirer = require('inquirer');
+const path = require('path');
+
+let appDefinitionSettings;
```
jelz, line 4: Why do we keep it around in a stateful manner here?

template/README.md
```diff
@@ -0,0 +1,28 @@
+This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
```
jelz, line 1: I'd say it was bootstrapped with `create-contentful-app`, mentioning CRA in "Learn More" is enough

template/README.md
```diff
@@ -0,0 +1,28 @@
...
+It correctly bundles React in production mode and optimizes the build for the best performance.
+
+The build is minified and the filenames include the hashes.
+Your app is ready to be deployed!
+
+## Learn More
+
+Create Contentful App uses [Create React App](https://create-react-app.dev/). You can learn more in the [Create React App documentation](https://facebook.github.io/create-react-app/docs/getting-started) and how to further customize your app.
+
+To learn React, check out the [React documentation](https://reactjs.org/).
```
jelz, line 28: Skip

lib/index.js
kdamball: I must've missed some changes I made last evening when doing the folder migration 🤦

lib/prompt-app-definition.js
kdamball: Fixed - was probably because of the previous /bin setup

.eslintignore
kdamball: Remnant of the previous logic - removed now

lib/prompt-app-definition.js
```diff
@@ -0,0 +1,58 @@
+const inquirer = require('inquirer');
+const path = require('path');
+
+module.exports = async function promptAppDefinition() {
+  let appDefinitionSettings = await inquirer.prompt([
```
jelz, line 5: `const`

kdamball: while we can use const here, I used let because we're modifying locations property a few lines later.

jelz: I think we should always use `const` if the reference doesn't change

PR #4, opened 2020-07-14
chore: formatting & copy updates

fabe:
This updates the formatting and copy. I also `tildify` the paths:

image
image

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,28 +2,33 @@
 
 /* eslint-disable no-console, no-process-exit */
 
+const chalk = require('chalk');
 const spawn = require('cross-spawn');
 const path = require('path');
 const Spinner = require('cli-spinner').Spinner;
 const promptAppDefinitionSettings = require('./prompt-app-definition');
 const createAppDefinition = require('./create-app-definition');
+const readline = require('readline');
+const stream = process.stdout;
 
 const command = process.argv[2];
 const appFolder = process.argv[3];
 
-const mainCommand = 'npx create-contentful-app';
+const localCommand = 'create-contentful-app';
+const mainCommand = `npx ${localCommand}`;
 
 function onSuccess(spinner, folder) {
   spinner.stop(true);
 
-  console.log();
-  console.log(`Success! Created ./${folder}`);
-  console.log();
-  console.log(`We suggest that you begin by running:`);
-  console.log();
-  console.log(`   cd ${folder}`);
-  console.log(`   ${mainCommand} create-definition`);
-  console.log();
+  console.log(`
+${chalk.cyan('Success!')} Created a new Contentful app in ${chalk.bold(`./${folder}`)}.
+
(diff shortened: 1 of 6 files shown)


Jwhiles:
are we publishing this under the contentful org? If so, I think we should also update the commands to `npx @contentful/create-contentful-app ...`

lib/login.js
```diff
@@ -33,7 +34,7 @@ module.exports = async function login() {
 
     return mgmtToken;
   } catch (err) {
-    console.log('Failed to login into Contentful');
+    console.log(chalk.red('Failed to login into Contentful.'));
```
kdamball, line 37: Is red going to look in dark editors?

fabe: It's using the `red` of the theme, so it'll be readable. Chalk handles everything.

Jwhiles, line 37: I'd prefer if we could change this to something like

```suggestion
    console.log(`${chalk.red('error:')} Failed to login into Contentful.'));
```
Output printed entirely in red is very hard to see on terminals with dark backgrounds

fabe:
> are we publishing this under the contentful org? If so, I think we should also update the commands to `npx @contentful/create-contentful-app ...`

Yes, we should update the readme once we have the package without `@contentful`!

lib/prompt-app-definition.js
```diff
@@ -1,16 +1,24 @@
...
   const appDefinitionSettings = await inquirer.prompt([
     {
       name: 'name',
       message: `App name (${path.basename(process.cwd())}):`
     },
     {
       name: 'locations',
-      message:
-        'Select where your app can be rendered. \n NOTE: Read more about app locations at https://ctfl.io/app-locations\n',
+      message: `Select where your app can be rendered in:`,
```
Jwhiles, line 21: "can be rendered in" sounds  strange to me. I think "can be rendered" would be fine.

lib/login.js
fabe: Good point, I'll do that for all error messages.

lib/index.js
```diff
@@ -2,42 +2,61 @@
...
-  console.log(`   ${mainCommand} create-definition`);
-  console.log();
+  console.log(`
+${chalk.cyan('Success!')} Created a new Contentful app in ${chalk.bold(
+    tildify(path.resolve(process.cwd(), folder))
+  )}.
+
+We suggest that you begin by running:
+
+    ${chalk.cyan(`cd ${folder}`)}
```
kdamball, line 31: Same thing here with the colours. Can we make sure it doesn't conflict? I think the `dim` stuff you used somewhere might be the way to go :)

fabe: It's using whatever `cyan` color is specified for the current theme, so the actual color will differ between themes. Our `inquirer.prompt` is also using `chalk.cyan` behind the scenes.

PR #5, opened 2020-07-14
feat(rename): Remove org prefix

kdamball:
From @fabe recommendation


fabe:
We can also remove it in the readme!

PR #9, opened 2020-07-15
chore: Minor clean up on docs and adding github link to npm

kdamball:


fabe:
Can we also include the docs link in the template?

kdamball:
@fabe does it make sense to include it there considering the video/link seem to help someone before they've bootstrapped while the template is everything after?

fabe:
@kdamball It gives more context about the whole thing and what happened, especially also about the `create-definition` part which happens after the app is bootstrapped.

kdamball:
@fabe added it

PR #10, opened 2020-07-16
fix: Use npm template package

kdamball:
As opposed to using a local one with [`file:` option](https://create-react-app.dev/docs/custom-templates/#testing-a-template)

### Further Details:

https://github.com/facebook/create-react-app/blob/v3.4.1/packages/create-react-app/createReactApp.js#L750 is the problem. It would parse `@contentful` in our name and break. It's partially also a bug on create-react-app. If a template name has `@` in it's name, it will break. Unless it's part of the organization name.

We already published a template on npm: https://www.npmjs.com/package/@contentful/cra-template-create-contentful-app. 

FYI, this means going forward we will publish 3 times: once for template, for org scoped pkg and create-contentful-app parked name.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -7,9 +7,9 @@ const spawn = require('cross-spawn');
 const path = require('path');
 const tildify = require('tildify');
 const Spinner = require('cli-spinner').Spinner;
+const readline = require('readline');
 const promptAppDefinitionSettings = require('./prompt-app-definition');
 const createAppDefinition = require('./create-app-definition');
-const readline = require('readline');
 const stream = process.stdout;
 
 const command = process.argv[2];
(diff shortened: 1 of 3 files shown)


lib/index.js
```diff
@@ -7,9 +7,10 @@ const spawn = require('cross-spawn');
 const path = require('path');
 const tildify = require('tildify');
 const Spinner = require('cli-spinner').Spinner;
+const readline = require('readline');
+const createReactApp = require.resolve('create-react-app');
```
jelz, line 11: can we move it to the function? It's not actual require of a module, I'd put it next to args so we don't mix things

kdamball:
@Jwhiles while using `node` to run CRA - it's unavoidable. We can use `npx` but someone brought up an issue with running `npx` inside parent process of `npx`.

jelz:
@Jwhiles should work with `npx` as a subprocess but it is fragile to me. With `resolve` you're guaranteed to run a direct dependency of CCA (run with `npx` too) rather than allowing the child `npx` to decide on its own. While debugging we've seen the child `npx` download deps of CRA again.

PR #11, opened 2020-07-26
fix: open login.js oauth url in linux and windows

ja-martinez:
I was having trouble using this CLI because I'm on Ubuntu and the module was trying to open the OAuth link using the "open" with the spawn module. The problem is that "open" only works in mac os terminals. I fixed this on my end by borrowing the openBrowser function that create-react-app uses to open links. I hope this can be of some help!


lib/openBrowser.js
```diff
@@ -0,0 +1,98 @@
...
+  } else {
+    action = Actions.BROWSER;
+  }
+  return { action, value, args };
+}
+
+function executeNodeScript(scriptPath, url) {
+  const extraArgs = process.argv.slice(2);
+  const child = spawn(process.execPath, [scriptPath, ...extraArgs, url], {
+    stdio: 'inherit'
```
kdamball, line 40: Please change this to `silent: true`. We shouldn't include any output here as we dont do anything with it. Maybe instead log error in case of failure :)

lib/openBrowser.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+
+/**
+ * Reads the BROWSER environment variable and decides what to do with it. Returns
+ * true if it opened a browser or ran a node.js script, otherwise false.
+ */
+function openBrowser(url) {
+  const { action, value, args } = getBrowserEnv();
+  switch (action) {
```
kdamball, line 85: We currently don't support `BROWSER` env variable. Ideally, we can simplify this to only start the browser process using `open` module. So, maybe change this PR to use `open` node module instead of the built-in process. And maybe throw an error in case open fails.

kdamball (approved):
Thanks!

Jwhiles:
I've taken a look at this, and verified that it works and is the same approach used in Create-React-App - so happy to merge it :)

Jwhiles (approved):

PR #16, opened 2020-08-13
chore: introduce local dev mode

shikaan:
## Context

With the current setup we aren't able to test local changes to templates. This change aims to make this possible by means of an env variable

> ⚠️ Warning
> This is just a small idea which I tested locally, but might needs some more thought. Scope of this PR is mainly to get some feedback and work a good solution together

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,8 @@
   "scripts": {
     "commit": "git-cz",
     "prettier": "prettier **/*.{js,jsx,ts,tsx} scripts/**/*.js --write --ignore-path .gitignore",
-    "commitmsg": "validate-commit-msg"
+    "commitmsg": "validate-commit-msg",
+    "start": "MODE=local node ./lib/index.js"
   },
   "dependencies": {
     "chalk": "^4.1.0",
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "commit": "git-cz",
     "prettier": "prettier **/*.{js,jsx,ts,tsx} scripts/**/*.js --write --ignore-path .gitignore",
     "commitmsg": "validate-commit-msg",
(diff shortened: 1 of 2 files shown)


davidfateh:
If this is a package that gets installed to your global npm, why not just use `npm link` when developing locally?

package.json
```diff
@@ -48,7 +48,8 @@
   "scripts": {
     "commit": "git-cz",
     "prettier": "prettier **/*.{js,jsx,ts,tsx} scripts/**/*.js --write --ignore-path .gitignore",
-    "commitmsg": "validate-commit-msg"
+    "commitmsg": "validate-commit-msg",
+    "start": "MODE=local node ./lib/index.js"
```
kdamball, line 52: Is there a reason for this command? As opposed to the regular `npx create-contentful-app`?

shikaan:
@davidfateh That won't solve this problem.

The issue we're facing here is that we're sourcing templates from another npm package (namely `@contentful/cra-template-create-contentful-app`). So updates to `@contentful/create-contentful-app` won't show up here, unless you hard-wire local templates.

Check [this](https://create-react-app.dev/docs/custom-templates/#testing-a-template) and [this](https://github.com/contentful/create-contentful-app/pull/16/files#diff-6d186b954a58d5bb740f73d84fe39073R65) for reference

package.json
shikaan: It's mainly a utility script to not forget to add the `MODE=local` whenever you develop locally. It was just the first idea off the top of my head, any other suggestions would be appreciated

kdamball: Maybe `npm run dev`?

shikaan: Pushed another commit with this change

PR #17, opened 2020-08-13
chore: add simple deploy script

shikaan:
## Context

This script is meant to automate the current madness we go through to deploy this package. 

Nothing particularly fancy, and probably this cannot be merged before we produce documentation for these steps, but I have just used it for last release and seems to be working. We can get back to it whenever we are rolling out next release

### DIFF
--- a/scripts/deploy.js
+++ b/scripts/deploy.js
@@ -1,0 +1,32 @@
+#! /usr/bin/env node
+
+const fs = require('fs')
+const childProcess = require('child_process')
+
+const PACKAGES = ['@contentful/create-contentful-app', 'create-contentful-app', '@contentful/cra-template-create-contentful-app']
+const ORIGINAL_PACKAGE_JSON = require('../package.json')
+
+if (!process.env.NPM_TOKEN) {
+  throw new Error('Missing NPM_TOKEN!')
+}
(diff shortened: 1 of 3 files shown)


scripts/deploy.js
```diff
@@ -0,0 +1,32 @@
...
+  const packageJson = {...ORIGINAL_PACKAGE_JSON, name: package}
+
+  console.log(` > 📝 Updating package.json with name: ${package}...`)
+  fs.writeFileSync('../package.json', JSON.stringify(packageJson))
+
+  console.log(` > ⚙️  Updating package-lock.json by means of "npm i"...`)
+  childProcess.execSync('npm i', {cwd: '../'})
+
+  console.log(` > 📚 Publishing on the registry...`)
+  childProcess.execSync('npm publish', {cwd: '../'})
```
kdamball, line 25: Is it piping things like 2-factor auth code entry back to the original shell?

kdamball:
🙏

scripts/deploy.js
shikaan: Uhm, I need to double check actually. I think I have inserted my code soon before, so it didn't ask and went through

PR #21, opened 2020-08-21
chore: [EXT-2124] automate dependecy check

shikaan:
## Context

This change is mean to provide the capability to update automatically dependencies in `template.json`.

Follows up from #18 

### Key differences with agreed approach

* **not using `init`**
  To perform `npm outdated` check we don't need the packages to be installed. Plus, creating a `package.json` from the template is super easy and this saves us a bunch of minutes of CI time

### Points to discuss

* **multiple dependencies**
  As we have agreed, this automation updates all the packages in the `template.json`. We might want to have fine grained control over which dependency gets updated and when.
* **ignored packages**
  I have added a list of packages we might want to ignore while performing this check. Currently is empty. Do you think it's something useful?

### To Dos

- [x] install updated dependencies, run tests and push the test report as the pull request description
- [x] cleanup stale PRs when creating new ones

---

I would like to get your opinion on the points to discuss before proceeding with the rest of the to-dos

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ jobs:
       - run:
           name: Check Outdated
           working_directory: ./dependency-check
-          command: (npm outdated --json || true ) > ./outdated.json
+          command: (npm outdated --json || true) > ./outdated.json
       - run:
           name: Create new template
           command: node ./scripts/dependency-check/create-template-from-outdated
@@ -35,7 +35,7 @@ jobs:
             cd "${TMPDIR}/${PROJECT}"
             npx @contentful/create-contentful-app init test
             cd test
-            ( npm t -- --ci --silent || true ) > "${FORMER_CWD}/dependency-check/test-report"
+            (CI=true npm t || true) 1>"${FORMER_CWD}/dependency-check/test-report" 2>&1
       - run:
           name: Create Pull Request
           command: npm install @octokit/rest@18 simple-git@2 && node ./scripts/dependency-check/create-pull-request
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,16 @@ jobs:
           name: Create new template
           command: node ./scripts/dependency-check/create-template-from-outdated
       - run:
+          name: Run Tests and Produce Reports
+          command: |
+            FORMER_CWD=$(pwd)
+            PROJECT=test-$(date +%s)
+            mkdir "${TMPDIR}/${PROJECT}"
+            cd "${TMPDIR}/${PROJECT}"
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,59 @@
...
+          name: Check Outdated
+          working_directory: ./dependency-check
+          command: (npm outdated --json || true ) > ./outdated.json
+      - run:
+          name: Create new template
+          command: node ./scripts/dependency-check/create-template-from-outdated
+      - run:
+          name: Run Tests and Produce Reports
+          command: |
+            FORMER_CWD=$(pwd)
```
jelz, line 32: most likely you want quotes here. You can grab the whole script and run it through https://www.shellcheck.net/

.circleci/config.yml
```diff
@@ -0,0 +1,59 @@
...
+            cd "${TMPDIR}/${PROJECT}"
+            npx @contentful/create-contentful-app init test
+            cd test
+            npm t > "${FORMER_CWD}/dependency-check/test-report"
+            cat "${FORMER_CWD}/dependency-check/test-report"
+      - run:
+          name: Create Pull Request
+          command: echo 'PR' #npm install @octokit/rest simple-git && node ./scripts/dependency-check/create-pull-request
+      - run:
+          name: Cleanup
```
jelz, line 44: Is it needed?

scripts/dependency-check/create-pull-request
```diff
@@ -0,0 +1,54 @@
...
+  await gitClient
+    .addConfig('user.name', 'Dependency Checker')
+    .addConfig('user.email', 'dependency-checker@contentful.org')
+
+  await gitClient.checkoutBranch(branch, 'master')
+
+  await gitClient.add([TEMPLATE_PATH])
+
+  // TODO: 
+  //   if we decide to do this one dependency at time,
```
jelz, line 32: I actually think all bumps in PRs is better

.circleci/config.yml
```diff
@@ -0,0 +1,59 @@
...
+            PROJECT=test-$(date +%s)
+            mkdir "${TMPDIR}/${PROJECT}"
+            cd "${TMPDIR}/${PROJECT}"
+            npx @contentful/create-contentful-app init test
+            cd test
+            npm t > "${FORMER_CWD}/dependency-check/test-report"
+            cat "${FORMER_CWD}/dependency-check/test-report"
+      - run:
+          name: Create Pull Request
+          command: echo 'PR' #npm install @octokit/rest simple-git && node ./scripts/dependency-check/create-pull-request
```
jelz, line 42: If you do it like that you should install at specific version (for example `simple-git@2.5`)

PR #50, opened 2020-08-25
fix: don't auto close PRs

shikaan:
## Context

This is to prevent [this](https://github.com/contentful/create-contentful-app/pull/49):

* find old dependencies;
* create the PR;
* in the attempt to close old PRs it also closes the one just opened

### DIFF
--- a/scripts/dependency-check/create-pull-request
+++ b/scripts/dependency-check/create-pull-request
@@ -86,7 +86,7 @@ const OWNER = 'contentful';
   const { data: pullRequests } = await githubClient.pulls.list({ owner: OWNER, repo: REPOSITORY, state: 'open' });
 
   for (const pullRequest of pullRequests) {
-    if (pullRequest.title === PULL_REQUEST_TITLE) {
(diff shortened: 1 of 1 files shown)


davidfateh (approved):

scripts/dependency-check/create-pull-request
```diff
@@ -86,7 +86,7 @@ const OWNER = 'contentful';
   const { data: pullRequests } = await githubClient.pulls.list({ owner: OWNER, repo: REPOSITORY, state: 'open' });
 
   for (const pullRequest of pullRequests) {
-    if (pullRequest.title === PULL_REQUEST_TITLE) {
+    if (pullRequest.title === PULL_REQUEST_TITLE && pullRequest.number !== newPullRequest.number) {
```
kdamball, line 89: Why not use the number only? It should be a unique identifier on its own.

Jwhiles (approved):

scripts/dependency-check/create-pull-request
shikaan: Because we want to close all the PRs with the same title and different number, not all the PRs in general

kdamball (approved):

PR #63, opened 2020-09-03
feat: Add links to libraries

stefanjudis:
Here we go. :) @kdamball.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,14 @@
 # Create Contentful App
 
-This project will make it easy to bootstrap [Contentful Apps](https://www.contentful.com/developers/docs/extensibility/app-framework/), with [React](https://reactjs.org/),
-[Forma 36](https://f36.contentful.com/), and [Field Editors](https://www.contentful.com/developers/docs/extensibility/field-editors/).
+This project makes it easy to bootstrap [Contentful Apps](https://www.contentful.com/developers/docs/extensibility/app-framework/) with [React](https://reactjs.org/) and [Forma 36](https://f36.contentful.com/) (Contentful's design system).
 
 ## Requirements
+
 Node 8 or higher
 NPM 5.2 and higher
 
 # Quick Overview
+
 To start developing your app:
 
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -23,3 +25,10 @@ npx @contentful/create-contentful-app create-definition
 ```
 
 [Read more](https://www.contentful.com/developers/docs/extensibility/app-framework/create-contentful-app/) and check out the video on how to use the CLI.
+
+## Libraries to use
+
+To make your app look and feel like Contentful use the following libraries:
+
+- [Forma 36](https://f36.contentful.com/) – Contentful's design system
```
kdamball, line 33: We already have forma at the top :)

kdamball (commented):
I would actually include these only in https://github.com/contentful/create-contentful-app/blob/master/template/README.md if possible. Note that this is in the bootstrapped app's README :) wdyt @fabe

README.md
stefanjudis: 🤦 Sorry about that. :D

README.md
```diff
@@ -1,13 +1,14 @@
 # Create Contentful App
 
-This project will make it easy to bootstrap [Contentful Apps](https://www.contentful.com/developers/docs/extensibility/app-framework/), with [React](https://reactjs.org/),
-[Forma 36](https://f36.contentful.com/), and [Field Editors](https://www.contentful.com/developers/docs/extensibility/field-editors/).
+This project makes it easy to bootstrap [Contentful Apps](https://www.contentful.com/developers/docs/extensibility/app-framework/) with [React](https://reactjs.org/) and [Forma 36](https://f36.contentful.com/) (Contentful's design system).
```
stefanjudis, line 3: I removed the field editors because they're not included by default. :)

fabe (approved):

kdamball (commented):
@stefanjudis can you add a version bump from latest master and also run `npm i` to make sure lock file is also updated?

PR #131, opened 2020-11-02
feat: [EXT-2278] add warning notice page for local development

Jwhiles:
Adding a 'landing screen' 
image

### DIFF
--- a/template/public/index.html
+++ b/template/public/index.html
@@ -15,11 +15,5 @@
       To begin the development, run `npm start`.
       To create a production bundle, use `npm run build`.
     -->
-    <script>
-      // You can remove this script tag before deploying your app
-      if (window.self === window.top) {
-        window.location.replace('https://app.contentful.com/deeplink?link=apps');
-      }
-    </script>
   </body>
 </html>
--- a/template/src/components/LocalhostWarning.tsx
+++ b/template/src/components/LocalhostWarning.tsx
@@ -9,22 +9,18 @@ const LocalhostWarning = () => {
         justifyContent: 'center',
         marginTop: '40px'
       }}>
-      <Note title="It looks like you're running your app locally." style={{ maxWidth: '800px' }}>
+      <Note title="App running outside of Contentful" style={{ maxWidth: '800px' }}>
         <Paragraph>
-          If this is your first time running your app, please run `npx
-          @contentful/create-contentful-app create-defintion` in your terminal then{' '}
-          <TextLink href="https://app.contentful.com/deeplink?link=apps">
-            install your app in a Contentful space.
+          Contentful Apps need to run inside the Contentful web app to function properly. Install
+          the app into a space and render your app into one of the{' '}
+          <TextLink href="https://www.contentful.com/developers/docs/extensibility/ui-extensions/sdk-reference/#locations">
+            available locations
           </TextLink>
-        </Paragraph>
-
-        <br />
-
-        <Paragraph>
-          More information about local development of Contentful apps{' '}
-          <TextLink href="https://www.contentful.com/developers/docs/extensibility/ui-extensions/faq/#how-can-i-develop-with-the-ui-extension-sdk-locally">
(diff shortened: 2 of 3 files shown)


template/public/index.html
```diff
@@ -15,11 +15,5 @@
       To begin the development, run `npm start`.
       To create a production bundle, use `npm run build`.
     -->
-    <script>
```
kdamball, line 18: I missed this - why are we removing the redirect?

template/src/index.tsx
```diff
@@ -23,47 +23,55 @@ import Sidebar from './components/Sidebar';
 import Field from './components/Field';
 import Dialog from './components/Dialog';
 
-init((sdk: BaseExtensionSDK) => {
+import LocalhostWarning from './components/LocalhostWarning';
+
+if (process.env.NODE_ENV === 'development' && window.self === window.top) {
+  // You can remove this if block before deploying your app
   const root = document.getElementById('root');
```
kdamball, line 30: fwiw - this can go outside the if statement.

template/public/index.html
Jwhiles: In order to show the explanation warning screen :)

stefanjudis: We've seen several people in the community being confused by the redirect. First, they don't understand that they shouldn't open `localhost` directly and if they do, they're completely lost with the target of the redirect. :)

template/src/components/LocalhostWarning.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <div
+      style={{
+        display: 'flex',
+        justifyContent: 'center',
+        marginTop: '40px'
+      }}>
+      <Note title="It looks like you're running your app locally." style={{ maxWidth: '800px' }}>
+        <Paragraph>
+          If this is your first time running your app, please run `npx
+          @contentful/create-contentful-app create-defintion` in your terminal then{' '}
```
davidfateh, line 15: While I think this is a good approach, I still feel a bit hesitant about this appDefinition flow. I think for first-timers, this is still pretty confusing and the appDefinition might make more sense when they see the UI version of it.

That being said, we can always gather more feedback once this version is released to see if people are still confused by it.

davidfateh (approved):

template/src/components/LocalhostWarning.tsx
stefanjudis: ☝️ I agree with this. I'd favor to give two options here:

- a link to documentation explaining the concept and the steps in the ui
- a "alright, you know what ya doing" way promoting the cli command

template/src/index.tsx
```diff
@@ -23,47 +23,55 @@ import Sidebar from './components/Sidebar';
 import Field from './components/Field';
 import Dialog from './components/Dialog';
 
-init((sdk: BaseExtensionSDK) => {
+import LocalhostWarning from './components/LocalhostWarning';
+
+if (process.env.NODE_ENV === 'development' && window.self === window.top) {
```
kdamball, line 28: Do we care if we're not in Dev mode? i.e: before we'd redirect even when not in dev mode

kdamball: Otherwise - it looks good to me

template/src/index.tsx
```diff
@@ -23,47 +23,55 @@ import Sidebar from './components/Sidebar';
...
-  // doesn't require it
-  ComponentLocationSettings.forEach(componentLocationSetting => {
-    if (sdk.location.is(componentLocationSetting.location)) {
-      render(componentLocationSetting.component, root);
-    }
+    // Select a component depending on a location in which the app is rendered.
+    //
+    // NB: Location "app-config" is auto-included in the list as most apps need it
+    // You can remove it (and on the app definition also) in case the app
+    // doesn't require it
```
VaguelySerious, line 70: How is it auto included? It seems to be just an element in the array like any other location

PR #134, opened 2020-11-03
Fix: enable node 15

Jwhiles:
CCA was crashing when used with Node-15. It appears to have been a dependency resolution issue. I've regenerated the lockfile with node-14, and it now appears to work. 

I've tested with node 10, 12, 14, and 15.

I've also bumped our nvmrc to v14, to keep the lockfile in this state.


andipaetzold:
Should immediately go to Node 12 like in create-contentful-extension? https://github.com/contentful/create-contentful-extension/pull/89

kdamball (commented):
`npm i` to make this deployment ready

Jwhiles:
> `npm i` to make this deployment ready. And maybe change `engine: {node: >=10}` in `package.json` to something more avoid v15?

I'm not sure what you mean about running `npm i`? I've already regenerated the lock file.

I don't really see a reason to exclude node 15 - as it seems to work once we generate the lockfile with node 12

kdamball:
I meant bumping the version in `package.json` and then running `npm i` - or manually changing package-lock

andipaetzold (approved):

PR #135, opened 2020-11-03
fix: don't open browser on start

Jwhiles:
This was confusing people, as the ui-extension-sdk doesn't work
unless run in an iFrame


andipaetzold (approved):
This also confused me 👍

fabe (approved):
boom goes the dynamite

PR #141, opened 2020-11-27
feat/introduce dependabot

shikaan:
## Context

This change is meant to introduce dependabot. However, given that we want to keep the dependencies automatically up to date, we do need some testing for the CLI tool to make sure nothing breaks. Therefore this PR also includes some unit tests for the cli tool itself

### DIFF
--- a/lib/create-app-definition.js
+++ b/lib/create-app-definition.js
@@ -3,7 +3,7 @@
 const chalk = require('chalk');
 const inquirer = require('inquirer');
 const path = require('path');
-const { createClient } = require('contentful-management');
+const cma = require('contentful-management');
 const login = require('./login');
 
 async function fetchOrganizations(client) {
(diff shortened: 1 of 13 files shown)


lib/create-app-definition.js
```diff
@@ -3,7 +3,7 @@
 const chalk = require('chalk');
 const inquirer = require('inquirer');
 const path = require('path');
-const { createClient } = require('contentful-management');
+const cma = require('contentful-management');
```
shikaan, line 6: This has been introduced to allow proxyquire to do its job for testing

lib/create-app-definition.js
```diff
@@ -23,75 +23,79 @@ ${chalk.red(
 ${err.message}
     `);
 
-    process.exit(1);
+    throw err;
```
shikaan, line 26: This now throws and the exist happens once at the end of the main function block. This was namely to make it testable

Jwhiles (approved):
Looks great!

Issue #160, opened 2020-11-30
New app doesn't build


caspardue:
Hi,

I'm getting an error when trying to build.

Here's what I did:
`npx @contentful/create-contentful-app init my-first-app`
`npm run build`

The error I'm getting:

```
> react-scripts build

Creating an optimized production build...
Failed to compile.

./node_modules/@contentful/forma-36-react-components/dist/styles.css
ParserError: Syntax Error at line: 1, column 23
```

Bonus issue:

When running `npm run start` on Windows I get:

```
$ BROWSER=none react-scripts start
'BROWSER' is not recognized as an internal or external command,
operable program or batch file.
```

This is because setting environment variables on Windows normally need a package like [cross-env](https://www.npmjs.com/package/cross-env]) to work.

caspardue:
Related issue https://github.com/facebook/create-react-app/issues/10130

PR #163, opened 2020-12-01
fix: use cross-env in start command

Jwhiles:
The script failed on most windows shells. Cross-env ensures that it will
work across platforms.

This was raised in #160


kdamball (approved):
Can you also bump forma to `3.68.10` and package.json so it fixes both

Jwhiles:
> Can you also bump forma to `3.68.10` and package.json so it fixes both

I'll make a separate PR

PR #174, opened 2020-12-07
Improve CI tests

andipaetzold:
I started this PR because node 15 currently doesn't fully work.

There are multiple issues in the CI config:
- `test-built-app` only tests current master and not the current
  revision
- node 15 is not tested
- Multiple node versions add complexity to the ci config

I use the official node orb as it makes life easier.

With this PR I slightly refactor the config and use a parameter matrix
to run jobs with multiple node versions. Also, the tests always use the
current revision and not what is currently deployed on npm.
Also, I combined both tests to avoid having to install dependencies
twice.

## ToDo:

- [ ] We currently don't specify the npm version which means that we don't test npm 7. Probably we should?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,10 +6,10 @@ orbs:
 jobs:
   dependency-check:
     docker:
-      - image: node:12
-        environment:
-          NPM_CONFIG_PROGRESS: 'false'
-          NPM_CONFIG_LOGLEVEL: warn
+       - image: node:12
+         environment:
+            NPM_CONFIG_PROGRESS: 'false'
+            NPM_CONFIG_LOGLEVEL: warn
     steps:
       - checkout
       - run:
@@ -43,8 +43,8 @@ jobs:
           command: rm -rf ./dependency-check
 
   lint-and-test:
-    executor:
-      name: node/default
+    docker:
+       - image: node:15
     parameters:
       node-version:
         type: string
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,23 +1,15 @@
 version: 2.1
 
-environment: &base-env
-    NPM_CONFIG_PROGRESS: 'false'
-    NPM_CONFIG_LOGLEVEL: warn
-    
-executors:
-  node-container:
-    docker:
-      - image: node:12
-        environment: *base-env
-
-  node-container-next-lts:
-    docker:
-      - image: node:14
-        environment: *base-env
+orbs:
+  node: circleci/node@4.1.0
 
 jobs:
   dependency-check:
-    executor: node-container
+    docker:
+      - image: node:12
+        environment:
+          NPM_CONFIG_PROGRESS: 'false'
+          NPM_CONFIG_LOGLEVEL: warn
     steps:
       - checkout
       - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -62,31 +61,18 @@ jobs:
...
-    executor: node-container
-    steps:
-      - checkout
+          command: npm test
       - run:
-          name: Run Tests
+          name: Build app and run its tests
           environment:
-            MODE: 'ci'
+            MODE: 'local'
```
shikaan, line 68: This needs to be 'ci', we don't want to test a new template here (since we are not creating a new template anyway). Tests for the new template are run in the other job (dependency-check), whose environment is set in Circle CI

.circleci/config.yml
```diff
@@ -62,31 +61,18 @@ jobs:
...
-            npm t
-  
-  test-built-app-next-lts:
-    <<: *test-built-app
-    executor: node-container-next-lts
+            node lib/index.js init $PROJECT
+            cd $PROJECT
+            npm test
+            cd ..
+            rm -r $PROJECT
```
shikaan, line 75: If the test fails, you would have the workspace littered. This is why the temp folder was used before.

Also, we want to use "what we ship", rather than "what we think we ship" hence why we were using the `npx` command there

.circleci/config.yml
```diff
@@ -1,23 +1,15 @@
...
+  node: circleci/node@4.1.0
 
 jobs:
   dependency-check:
-    executor: node-container
+    docker:
+       - image: node:12
+         environment:
+            NPM_CONFIG_PROGRESS: 'false'
+            NPM_CONFIG_LOGLEVEL: warn
```
shikaan, line 12: If you want to use the orb, I would sill attach this environment to the orb and use it across the board. Otherwise we get inconsistent result among different runs

.circleci/config.yml
```diff
@@ -50,10 +42,17 @@ jobs:
           name: Cleanup
           command: rm -rf ./dependency-check
 
-  lint-and-test: &lint-and-test
-    executor: node-container
+  lint-and-test:
+    docker:
+       - image: node:15
```
shikaan, line 47: Node 15 is not LTS, it should not be the base

shikaan (commented):
I would have added another executor block in the configuration, to run the tests on Node 15 (although it's not LTS, so it's a can of worms I would not open either way). 

Also take another look to the rest of the changes as some of them seem to be dangerous

```
  node-container-latest:	
    docker:	
      - image: node:15	
        environment: *base-env
```

.circleci/config.yml
andipaetzold: Agree - this was leftover when testing around with installing different npm versions

.circleci/config.yml
andipaetzold: What do you mean with "inconsistent"? This only changes the logs, right?

.circleci/config.yml
andipaetzold: Okay, I wasn't aware of that behavior of `npx`. From simply reading the code it looks like we would always run use the release version.

PR #218, opened 2021-01-05
feat: use @contentful/app-sdk

andipaetzold:
Replace `contentful-ui-extensions-sdk` with `@contentful/app-sdk`.
Both packages are identical but the App SDK should be used for apps.


kdamball (approved):
🎉

PR #235, opened 2021-01-15
feat: directly forward logs from npx process

andipaetzold:
We currently use `.stderr.on('data', [...])` and `.stdout.on('data', [...])` and put the data in `console.log`. That removes color and adds additional new lines

By using `{ stdio: 'inherit' }` we can forward the data directly to stdout.
> Pass through the corresponding stdio stream to/from the parent process. 

That way
- we keep the log color
- we don't add additional empty lines
- the output is prettier and easier to read

# Before
## Success

Screenshot 2021-01-15 at 14 57 59
Screenshot 2021-01-15 at 14 58 18

## Error

Screenshot 2021-01-15 at 14 56 02

# After
## Success

Screenshot 2021-01-15 at 14 54 59

## Error

Screenshot 2021-01-15 at 14 55 39

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -49,23 +49,12 @@ function initProject() {
 
     const args = [createReactApp, appFolder, '--template', templatePkg, '--use-npm'];
 
-    // start creating app
-    const appCreateProcess = spawn(initCommand, args);
-
-    appCreateProcess.stdout.on('data', data => {
-      console.log(data.toString());
-    });
-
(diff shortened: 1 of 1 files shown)


kdamball (approved):
Looks good but I'm curious why we didn't do this last time. I might be misremembering and confusing it with somewhere else where logging couldn't be done this way; @Jwhiles might know more

andipaetzold:
Just checked and the options were already available in older Node.js versions

PR #236, opened 2021-01-15
feat: Clarify supported node/npm versions

kdamball:
Still doesn't warn when using `npx` though

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,8 @@ This project makes it easy to bootstrap [Contentful Apps](https://www.contentful
 
 ## Requirements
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -4,7 +4,8 @@ This project makes it easy to bootstrap [Contentful Apps](https://www.contentful
 
 ## Requirements
 
-Node.js 12 or higher
+- Node.js v12 to v14
```
andipaetzold, line 7: Let's say v12 or v14.
v13 is not LTS and not developed anymore

andipaetzold (approved):

PR #356, opened 2021-03-26
chore: auto merge dep-check PRs [EXT-2537]

andipaetzold:
### 2. Automatically merges dep-check PRs before creating a new PR.

Unfortunately, it is currently not possible to enable the GitHub auto-merge feature via the API (yet). So, we need to do this manually for now. The new script fetches all open PRs, filters them to get the latest dep-check PR, verifies that all checks were successful, squashes that PR and then fetches the latest `master` to the local git. Fetching the latest `master` ensures that the new PR does not contain the same updates as the already merged PR.

It might make sense to run this in a separate job as the final pull might not include the previously merged PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,6 +60,12 @@ jobs:
     executor: node-container-lts
     steps:
       - checkout
+      - run: 
+          name: Install additional dependencies
+          command: npm install @octokit/rest@18 simple-git@2
+      - run:
+          name: Merge Pull Request
+          command: node ./scripts/dependency-check/merge-pull-request.js
       - run:
           name: Create Dependency Check sandbox
           command: mkdir -p ./dependency-check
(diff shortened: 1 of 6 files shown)


kdamball (approved):
As long as we make sure the newly created PR is from latest master after the "merge PRs" job is ran.

andipaetzold:
> As long as we make sure the newly created PR is from latest master after the "merge PRs" job is ran.

It's not explicitly stated in the API, but I assume that the merge is done and can be pulled after the request returned successfully.

https://docs.github.com/en/rest/reference/repos#merging

Alternatively, we could just use gitclient itself to merge. Might be the safer option.

andipaetzold:
Okay, the PR merge request seems to return when the PR is actually merged, so it is immediately available via git

PR #368, opened 2021-03-31
chore: switch to lerna

shikaan:
* switch project to lerna
* uses a `_` version of `create-contentful-app` to allow publishing under two names
* releases automatically upon merging to master (when test passes)
* changelogs will be recreated in the individual folders

I might need to make it a `fix` version (rather than a `chore`) to test the automatic deployment upon merging to master


packages/create-contentful-app/lib/index.js
```diff
@@ -7,7 +7,7 @@ const spawn = require('cross-spawn');
 const path = require('path');
 const tildify = require('tildify');
 const promptAppDefinitionSettings = require('./prompt-app-definition');
-const createAppDefinition = require('./create-app-definition');
+const createAppDefinition = require('./index');
```
andipaetzold, line 10: Does this work? This file requires itself.

packages/_create-contentful-app/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "@contentful/create-contentful-app": "^0.4.2"
+  },
+  "bugs": {
+    "url": "https://github.com/contentful/create-contentful-app/issues"
+  },
+  "homepage": "https://github.com/contentful/create-contentful-app#readme",
+  "directories": {
+    "lib": "lib",
+    "test": "test"
+  },
```
andipaetzold, line 55: Why do we need this?

package.json
```diff
@@ -1,70 +1,19 @@
 {
   "name": "@contentful/create-contentful-app",
-  "version": "0.4.2",
```
andipaetzold, line 3: This should have `"private": true` to avoid accidental publishing. Also, `version` is required (I think).

.circleci/config.yml
```diff
@@ -26,23 +26,23 @@ commands:
...
     steps:
       - checkout
       - run:
           name: Install Dependencies
-          command: npm ci
+          command: npm ci && npm run bootstrap
       - run:
           name: Link package
-          command: npm link
+          command: cd packages/create-contentful-app && npm link
```
andipaetzold, line 45: There is `working_directory` in Circle CI: https://circleci.com/docs/2.0/configuration-reference/#run

packages/_create-contentful-app/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "files": [
+    "lib"
```
andipaetzold, line 57:
```suggestion
    "index.js"
```

packages/_create-contentful-app/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "node": ">=12 <15",
+    "npm": ">=6 <7"
+  },
+  "main": "index.js",
+  "config": {
+    "commitizen": {
+      "path": "node_modules/cz-customizable"
+    },
+    "cz-customizable": {
+      "config": "cz-customizable.config.js"
```
andipaetzold, line 22: We could remove this as `cz-customizable.config.js` doesn't exist anyway and this is broken anyway.

Same applies for other package.json files

andipaetzold (changes requested):

andipaetzold:
Having `_create-contentful-app` publish `create-contentful-app` and `create-contentful-app` publishing `@contentful/create-contentful-app` confuses me. Can we somehow improve this.

Maybe have folders names `create-contentful-app` and `@contentful-create-contentful-app`? 🤔

packages/_create-contentful-app/package.json
shikaan: When you create a package with lerna you get it ootb, not necessarily required

package.json
shikaan: Version is taken from outside (`lerna.json`) and the private switch does not really matter, as publishing is now done by ci

.circleci/config.yml
shikaan: Thanks for the catch -.- It's a leftover

package.json
andipaetzold: But if you run `npm publish`, the regular `npm` publishing still gets triggered, right?

shikaan:
> Having `_create-contentful-app` publish `create-contentful-app` and `create-contentful-app` publishing `@contentful/create-contentful-app` confuses me. Can we somehow improve this.
> 
> Maybe have folders names `create-contentful-app` and `@contentful-create-contentful-app`? 🤔

I thought the same initially, but in reality the outlier is the unprefixed package. Are we okay with prefixing everything _but_ one folder? Looks too noisy to me

andipaetzold:
> > Having `_create-contentful-app` publish `create-contentful-app` and `create-contentful-app` publishing `@contentful/create-contentful-app` confuses me. Can we somehow improve this.
> > Maybe have folders names `create-contentful-app` and `@contentful-create-contentful-app`? 🤔
> 
> I thought the same initially, but in reality the outlier is the unprefixed package. Are we okay with prefixing everything _but_ one folder? Looks too noisy to me

Probably fine to keep it as we only have three folders. But it initially confused me...

The template package shouldn't bet the `_` prefix.

shikaan:
@andipaetzold 

> Probably fine to keep it as we only have three folders. But it initially confused me...

Sounds good to me, not super opinionated

> The template package shouldn't bet the `_` prefix.

What do you mean?

packages/@contentful--create-contentful-app/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@contentful/create-contentful-app",
+  "version": "0.4.2",
+  "description": "A template for building Contentful Apps",
```
andipaetzold, line 4: This is not the template package

packages/@contentful--create-contentful-app/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "tildify": "^2.0.0"
+  },
+  "bugs": {
+    "url": "https://github.com/contentful/create-contentful-app/issues"
+  },
+  "homepage": "https://github.com/contentful/create-contentful-app#readme",
+  "directories": {
+    "lib": "lib",
+    "test": "test"
+  },
```
andipaetzold, line 53: Delete me

packages/@contentful--create-contentful-app/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@contentful/create-contentful-app",
+  "version": "0.4.2",
+  "description": "A template for building Contentful Apps",
+  "author": "Contentful GmbH",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/contentful/create-contentful-app.git",
```
andipaetzold, line 9:
```suggestion
    "url": "https://github.com/contentful/create-contentful-app.git",
```

packages/@contentful--cra-template-create-contentful-app/package.json
```diff
@@ -7,7 +7,7 @@
   "repository": {
     "type": "git",
     "url": "git+https://github.com/contentful/create-contentful-app.git",
```
andipaetzold, line 9: Should be `https://github.com/contentful/create-contentful-app.git`

packages/@contentful--create-contentful-app/package.json
```diff
@@ -0,0 +1,60 @@
+{
```
andipaetzold, line 1: This file is missing the keywords from the template package.json

andipaetzold (approved):
It's nitpick wednesday!

PR looks good. Just some package.json tweaking

kdamball:
Good start but I'm now wondering if there's a way to avoid duplicating a package folder just for avoiding the name parking? i.e: is there  a way to publish `packages/contentful--create-contentful-app` as both `create-contentful-app` and `@contentful/create-contentful-app`

kdamball (commented):
Good start but I'm now wondering if there's a way to avoid duplicating a package folder just for avoiding the name parking? i.e: is there a way to publish `packages/contentful--create-contentful-app` as both `create-contentful-app` and `@contentful/create-contentful-app`

shikaan:
> Good start but I'm now wondering if there's a way to avoid duplicating a package folder just for avoiding the name parking? i.e: is there a way to publish `packages/contentful--create-contentful-app` as both `create-contentful-app` and `@contentful/create-contentful-app`

@kdamball I have been looking at that, but I haven't been able to find any OOTB solution. Also, I think re-exporting the same package as the other one is easier to reason about: we won't need to introduce custom publishing logic (as we did before) and we're using standard publishing flow

PR #374, opened 2021-04-06
feat(scripts): extract script package

shikaan:


packages/@contentful--app-scripts/README.md
```diff
@@ -0,0 +1,77 @@
...
+> **:warning: Please note** 
+> 
+> Both interactive and nonInteractive version of the same script is meant to return the same result.
+
+### Create App Definition
+
+Allows creating a new [AppDefinition](https://www.contentful.com/developers/docs/extensibility/app-framework/app-definition/)
+provided a Content Management Token (more details [here](https://www.contentful.com/developers/docs/references/content-management-api/#/reference/personal-access-tokens)).
+
+It only runs in interactive mode.
```
andipaetzold, line 71: Why?

We should be able to create a new definition with a provided cma token.

packages/@contentful--create-contentful-app/package.json
```diff
@@ -28,13 +28,12 @@
...
     "lint": "eslint ./lib",
     "lint:fix": "npm run lint -- --fix",
-    "test": "mocha ./test/*.test.js",
-    "test:watch": "npm t -- --watch",
     "pre-commit": "lint-staged",
     "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
   },
   "dependencies": {
     "@contentful/cra-template-create-contentful-app": "^0.4.2",
+    "@contentful/app-scripts": "^0.4.2",
```
kdamball, line 36: Maybe wildcard this?

kdamball (approved):

packages/@contentful--app-scripts/README.md
```diff
@@ -0,0 +1,77 @@
...
+Scripts exported from this module will all be in the following shape
+
+```typescript
+interface Script<Result, Options> {
+  // query the user or local cache for required information
+  interactive: () => Result;
+  // run by automation (`--ci` flag), they would need all the information upfront
+  nonInteractive: (...options: Options) => Result;
+}
+```
```
andipaetzold, line 60: Do we really want to export an interactive API? I have never seen that before and also it feels wrong to me. As I user I never expect an API to show prompts to fill out missing options

packages/@contentful--app-scripts/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
+  },
+  "bugs": {
+    "url": "https://github.com/contentful/create-contentful-app/issues"
+  },
+  "homepage": "https://github.com/contentful/create-contentful-app#readme",
+  "directories": {
+    "lib": "lib",
+    "bin": "bin"
+  },
```
andipaetzold, line 43: 🤷  where do these come from?

packages/@contentful--app-scripts/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@contentful/app-scripts",
+  "version": "0.4.2",
+  "description": "A collection of scripts for building Contentful Apps",
+  "author": "Contentful GmbH",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/contentful/create-contentful-app.git",
+    "directory": "packages/app-scripts"
```
andipaetzold, line 10:
```suggestion
    "directory": "packages/@contentful--app-scripts"
```

packages/@contentful--app-scripts/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@contentful/app-scripts",
+  "version": "0.4.2",
+  "description": "A collection of scripts for building Contentful Apps",
+  "author": "Contentful GmbH",
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/contentful/create-contentful-app.git",
```
andipaetzold, line 9:
```suggestion
    "url": "https://github.com/contentful/create-contentful-app.git",
```

andipaetzold (changes requested):
Mainly looks good. Just some questions and minor fixes

packages/@contentful--app-scripts/README.md
shikaan: This was meant to reproduce existing functionality, which is interactive only. Do we have a case for non-interactive app definition creation?

Also, the implementation might not be as straightforward: you need to pass the rest of the details (locations, fields) and you either nest lots of flags, accept a JSON shaped object, or maybe something else. 

TL;DR: if we don't have a clear case for this, I wouldn't waste time and effort on it

packages/@contentful--app-scripts/README.md
shikaan: Maybe to plug it with your own flow: we're using standard tooling (`inquirer`, `commander`), it does not seem out of this world. 

OTOH, I am not super opinionated here and it felt like an easy win. Happy to tick it off

packages/@contentful--app-scripts/package.json
shikaan: Lerna 🙃 Also, it's compliant with [specification](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#directories) so I think it does not hurt

packages/@contentful--app-scripts/README.md
andipaetzold: Okay, I am happy to keep it that way.

packages/@contentful--app-scripts/package.json
andipaetzold: Aaaah. I was looking at the npm 7 specs: https://docs.npmjs.com/cli/v7/configuring-npm/package-json#directories
They removed all but `bin` and `man`.

packages/@contentful--app-scripts/README.md
andipaetzold: Maybe it would be good to get more opinions on this?

andipaetzold (approved):

PR #380, opened 2021-04-09
feat(scripts) upload bundle script

loweisz:
Adds the upload script to `app-scripts` in interactive and in non-interactive (`--ci`) mode.


packages/@contentful--app-scripts/lib/upload/index.js
```diff
@@ -0,0 +1,15 @@
...
+const { createAppBundle } = require('./create-app-bundle');
+
+const { buildAppUploadSettings } = require('./build-upload-settings');
+
+module.exports = {
+  async interactive() {
+    const settings = await buildAppUploadSettings();
+    await createAppBundle(settings);
+  },
+  async nonInteractive(program, options) {
```
shikaan, line 11: Why do we need `program` here?

packages/@contentful--app-scripts/lib/upload/get-upload-settings-args.js
```diff
@@ -0,0 +1,60 @@
...
+function validateArguments(options) {
+  Object.entries(requiredOptions).forEach(([option, argument]) => {
+    if (!options[option]) {
+      console.log(
+        `
+  ${chalk.red('Invalid Arguments:')} the argument ${chalk.cyan(argument)} was not defined
+  Run ${chalk.dim('contentful-app-scripts upload --help')} to see all required arguments
+        `
+      );
+      throw new Error('Invalid Arguments');
```
shikaan, line 24: nit: this might be a `TypeError`

packages/@contentful--app-scripts/lib/upload/create-zip-from-directory.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+async function createZipFileFromDirectory(path) {
+  try {
+    let zip = new JSZip();
+    const filePaths = await new Promise((res) => {
+      glob(path + '/**/*', { nodir: true }, (err, paths) => {
+        res(paths);
+      });
+    });
```
shikaan, line 35: You can use `util.promisify` here to stick with the async/await style we have elsewhere

packages/@contentful--app-scripts/lib/upload/create-zip-from-directory.js
```diff
@@ -0,0 +1,61 @@
...
+const fs = require('fs');
+const JSZip = require('jszip');
+const { showCreationError } = require('../utils');
+
+async function convertPathToFileData(filePath) {
+  return new Promise((res) => {
+    fs.readFile(filePath, 'utf8', (err, data) => {
+      res([filePath, data]);
+    });
+  });
```
shikaan, line 12: You can use `fs/promises` instead of wrapping. Although, being a script, also sync API would do

packages/@contentful--app-scripts/lib/upload/create-zip-from-directory.js
```diff
@@ -0,0 +1,61 @@
...
+      localPath = localPath.substring(localPath.length, path.length);
+    }
+    zip.file(localPath, data);
+  });
+  return zip;
+}
+
+async function createZipFileFromDirectory(path) {
+  try {
+    let zip = new JSZip();
```
shikaan, line 30: You can use [`adm-zip`](https://github.com/cthackers/adm-zip) here: it's MIT licensed and it allows adding a full folder by its path (so no need to do `fs` manipulation)

packages/@contentful--app-scripts/lib/upload/definition-api.js
```diff
@@ -0,0 +1,66 @@
+const chalk = require('chalk');
+const inquirer = require('inquirer');
+const ora = require('ora');
+
+async function fetchDefinitions(client, orgId) {
```
shikaan, line 5: Why do we duplicate this?

packages/@contentful--app-scripts/lib/upload/organization-api.js
```diff
@@ -0,0 +1,72 @@
+const inquirer = require('inquirer');
+const chalk = require('chalk');
+const ora = require('ora');
+
+async function fetchOrganizations(client) {
```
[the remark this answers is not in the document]
shikaan: Also using the error wrapper I was suggesting below and the plain client in CMA, this can become a couple of lines of code. You might not need this utility in the first place

packages/@contentful--app-scripts/lib/upload/organization-api.js
```diff
@@ -0,0 +1,72 @@
...
+  const { organizationId } = await inquirer.prompt([
+    {
+      name: 'organizationId',
+      message: 'Select an organization for your app:',
+      type: 'list',
+      choices: organizations,
+    },
+  ]);
+  return organizations.find((org) => org.value === organizationId);
+}
```
shikaan, line 47: Selecting from a list is something we do quite often. I would extract a utilty for it and reuse it for definitions and organizaions

packages/@contentful--app-scripts/lib/upload/definition-api.js
shikaan: Also, for the rest of the file, see comments on `organization-api`

packages/@contentful--app-scripts/lib/upload/index.js
shikaan, line 11: Why `program` here?

packages/@contentful--app-scripts/lib/upload/create-zip-from-directory.js
```diff
@@ -0,0 +1,61 @@
+const glob = require('glob');
+const chalk = require('chalk');
+const fs = require('fs');
+const JSZip = require('jszip');
+const { showCreationError } = require('../utils');
+
+async function convertPathToFileData(filePath) {
+  return new Promise((res) => {
```
shikaan, line 8: You can use `fs/promises` instead of wrapping. Although, being a script, also sync API would do

packages/@contentful--app-scripts/lib/upload/create-zip-from-directory.js
shikaan, line 30: You can use [`adm-zip`](https://github.com/cthackers/adm-zip) here: it's MIT licensed and it allows adding a full folder by its path (so no need to do `fs` manipulation)

packages/@contentful--app-scripts/lib/upload/create-zip-from-directory.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+    zip.file(localPath, data);
+  });
+  return zip;
+}
+
+async function createZipFileFromDirectory(path) {
+  try {
+    let zip = new JSZip();
+    const filePaths = await new Promise((res) => {
```
shikaan, line 31: You can use `util.promisify` here to stick with the async/await style we have elsewhere

shikaan (commented):
It's a good starting point. There are a couple fo things that can be streamlined and then we are ready to ship this :)

One question: what is it going to happen when we use the interactive mode _and_ we pass the parameters? Will it throw or will it just ask for them?

packages/@contentful--app-scripts/lib/upload/organization-api.js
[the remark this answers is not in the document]
loweisz: I actually already did that in the second PR about activation. But I moved it now here too

packages/@contentful--app-scripts/lib/upload/organization-api.js
```diff
@@ -0,0 +1,72 @@
...
+      value: org.sys.id,
+    }));
+  } catch (err) {
+    console.log(`
+${chalk.red(
+  'Error:'
+)} Could not fetch your organizations. Make sure you provided a valid access token.
+
+${err.message}
+    `);
```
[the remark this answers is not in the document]
loweisz: Good idea 👍

packages/@contentful--app-scripts/lib/upload/index.js
loweisz: I used it in a previous implementation, forgot to remove it

packages/@contentful--app-scripts/lib/upload/create-zip-from-directory.js
loweisz: it's not used anymore anyway :D

packages/@contentful--app-scripts/lib/upload/create-zip-from-directory.js
loweisz: that was a really good tip 😬  Could remove half of the code haha

packages/@contentful--app-scripts/lib/upload/organization-api.js
loweisz: nice idea I abstracted it in utils 👍

loweisz:
> It's a good starting point. There are a couple fo things that can be streamlined and then we are ready to ship this :)
> 
> One question: what is it going to happen when we use the interactive mode _and_ we pass the parameters? Will it throw or will it just ask for them?

@shikaan It actually will just ask for them and ignore the ones passed. I tried to make them conditional that when you pass `--ci` you will get a better notification that you have to pass them in that case. But somehow I couldn't find a good way 😬

But I think we could throw when passed arguments without `--ci`, do you think that would make more sense than just ignoring it?

shikaan:
> > It's a good starting point. There are a couple fo things that can be streamlined and then we are ready to ship this :)
> > One question: what is it going to happen when we use the interactive mode _and_ we pass the parameters? Will it throw or will it just ask for them?
> 
> @shikaan It actually will just ask for them and ignore the ones passed. I tried to make them conditional that when you pass `--ci` you will get a better notification that you have to pass them in that case. But somehow I couldn't find a good way 😬
> 
> But I think we could throw when passed arguments without `--ci`, do you think that would make more sense than just ignoring it?

I was more thinking about we don't ask for the things we already know. So, if something is passed (even in interactive mode), we don't request further details on that.

TL;DR: we keep the same interface for interactive and non-interactive commands, but in the interactive case the arguments are all optional. WDYT?

loweisz:
> > > It's a good starting point. There are a couple fo things that can be streamlined and then we are ready to ship this :)
> > > One question: what is it going to happen when we use the interactive mode _and_ we pass the parameters? Will it throw or will it just ask for them?
> > 
> > 
> > @shikaan It actually will just ask for them and ignore the ones passed. I tried to make them conditional that when you pass `--ci` you will get a better notification that you have to pass them in that case. But somehow I couldn't find a good way 😬
> > But I think we could throw when passed arguments without `--ci`, do you think that would make more sense than just ignoring it?
> 
> I was more thinking about we don't ask for the things we already know. So, if something is passed (even in interactive mode), we don't request further details on that.
> 
> TL;DR: we keep the same interface for interactive and non-interactive commands, but in the interactive case the arguments are all optional. WDYT?

Ahh I see ! That's a good idea, will look into it

packages/@contentful--app-scripts/lib/upload/create-app-bundle.js
```diff
@@ -0,0 +1,72 @@
...
+const ora = require('ora');
+const { showCreationError } = require('../utils');
+const { createClient } = require('contentful-management');
+
+const { createAppUpload } = require('./create-app-upload');
+
+async function createAppBundleFromUpload(settings, appUploadId) {
+  const clientSpinner = ora('Verifing your upload').start();
+  const client = createClient({ accessToken: settings.accessToken });
+  const org = await client.getOrganization(settings.organisation.value);
```
shikaan, line 11: nit: organi**z**ation for consistency

shikaan (commented):
Looks great 🎉  We only need to cover it with some tests and we're good to go 🚢

packages/@contentful--app-scripts/lib/upload/get-upload-settings-args.js
```diff
@@ -0,0 +1,60 @@
...
+  token: '--token',
+};
+
+function validateArguments(options) {
+  Object.entries(requiredOptions).forEach(([option, argument]) => {
+    if (!options[option]) {
+      console.log(
+        `
+  ${chalk.red('Invalid Arguments:')} the argument ${chalk.cyan(argument)} was not defined
+  Run ${chalk.dim('contentful-app-scripts upload --help')} to see all required arguments
```
andipaetzold, line 21: That command only works when the `contentful-app-scripts` are installed globally. If there is a chance that this error is displayed when using the CCA output, we should provide an alternative call here. Even `npx contentful-app-scripts --help` might do it.

Also: This should be `@contentful/app-scripts`

packages/@contentful--app-scripts/lib/upload/create-app-bundle.js
```diff
@@ -0,0 +1,72 @@
...
+  console.log(`
+  ${chalk.green(`NEXT STEPS:`)}
+
+    ${chalk.bold('You can activate this app bundle in your apps settings:')}
+
+      ${chalk.underline(`https://app.contentful.com/deeplink?link=app-definition-list`)}
+
+    ${chalk.bold('or by simply running the cli command:')}
+
+      ${chalk.underline(`contentful-app-scripts activate`)}
```
andipaetzold, line 69: Same issue with the app scripts package here.

packages/@contentful--app-scripts/bin/app-scripts
```diff
@@ -3,32 +3,41 @@
...
+    })
 
-    return command();
-  })
+  program
+    .command('upload')
+    .description('Upload your build folder and create an app bundle')
+    .option('--bundle-dir [directory]', 'The directory of your build folder')
+    .option('--organization-id [orgId]', 'The id of your organisation')
+    .option('--definition-id  [defId]', 'The id of your apps definition')
```
andipaetzold, line 31:
```suggestion
    .option('--definition-id [defId]', 'The id of your apps definition')
```

packages/@contentful--app-scripts/bin/app-scripts
```diff
@@ -3,32 +3,41 @@
 const { program } = require('commander');
 const { version } = require('../package.json');
 
-const {createAppDefinition} = require('../')
+const {createAppDefinition, upload} = require('../')
 
-function getCommand(program, command) {
+
+function runCommand(command, options) {
```
andipaetzold, line 9: This must be async and await for `nonInteractive()` and `interactive()`

packages/@contentful--app-scripts/bin/app-scripts
```diff
@@ -3,32 +3,41 @@
...
   program
-  .command('create-app-definition')
-  .description('Create a new AppDefinition for an App')
-  .action(() => {
-    const command = getCommand(program, createAppDefinition);
+    .command('create-app-definition')
+    .description('Create a new AppDefinition for an App')
+    .action((options) => {
+      runCommand(createAppDefinition, options);
+    })
```
andipaetzold, line 24: Should be an async function with await

packages/@contentful--app-scripts/bin/app-scripts
```diff
@@ -3,32 +3,41 @@
...
+    .command('upload')
+    .description('Upload your build folder and create an app bundle')
+    .option('--bundle-dir [directory]', 'The directory of your build folder')
+    .option('--organization-id [orgId]', 'The id of your organisation')
+    .option('--definition-id  [defId]', 'The id of your apps definition')
+    .option('--token [accessToken]', 'Your content management access token')
+    .option('--comment [comment]', 'Optional comment for the created bundle')
+    .action((options) => {
+      runCommand(upload, options);
+    })
```
andipaetzold, line 36: Should be an async function with await

andipaetzold (changes requested):
You alternate between using `module.exports = ` and `exports.xxx`. Any reason for that?
I'd make this consistent and always use `module.exports`

Also, whether a log line has a trailing period `.` is inconsistent.

packages/@contentful--app-scripts/lib/upload/create-app-bundle.js
```diff
@@ -9,8 +9,8 @@ async function createAppBundleFromUpload(settings, appUploadId) {
   console.log(settings.definition);
   const clientSpinner = ora('Verifing your upload').start();
   const client = createClient({ accessToken: settings.accessToken });
-  const org = await client.getOrganization(settings.organisation.value);
-  const appDefinition = await org.getAppDefinition(settings.definition.value);
+  const organization = await client.getOrganization(settings.organisation.value);
```
andipaetzold, line 12: Manuel was talking about `organisation` with an `s` here.

loweisz:
> You alternate between using `module.exports = ` and `exports.xxx`. Any reason for that?
> I'd make this consistent and always use `module.exports`
> 
> Also, whether a log line has a trailing period `.` is inconsistent.

Not really `exports` was used in the code base before. I changed it to `module.exports` now everywhere and also added a `.` to all logs 👍

packages/@contentful--app-scripts/lib/upload/get-upload-settings-args.js
loweisz: True ! I was planning to change that with the real command, changed it now

loweisz: True ! I was planning to change that with the real command, changed it now

packages/@contentful--app-scripts/lib/upload/create-app-bundle.js
```diff
@@ -0,0 +1,77 @@
...
+  }
+  bundleSpinner.stop();
+  return appBundle;
+}
+
+async function createAppBundleFromSettings(settings) {
+  let appUpload = null;
+  try {
+    appUpload = await createAppUpload(settings);
+    console.log('UPLOAD', appUpload);
```
shikaan, line 34: Is this leftover? 🤔

packages/@contentful--app-scripts/bin/app-scripts
```diff
@@ -3,32 +3,41 @@
...
   program
-  .command('create-app-definition')
-  .description('Create a new AppDefinition for an App')
-  .action(() => {
-    const command = getCommand(program, createAppDefinition);
+    .command('create-app-definition')
+    .description('Create a new AppDefinition for an App')
+    .action(async (options) => {
+      await runCommand(createAppDefinition, options);
+    })
```
shikaan, line 24: super-nit

```suggestion
    .action((options) => runCommand(createAppDefinition, options))
```

packages/@contentful--app-scripts/bin/app-scripts
```diff
@@ -3,32 +3,41 @@
...
-const {createAppDefinition} = require('../')
+const {createAppDefinition, upload} = require('../')
 
-function getCommand(program, command) {
+
+async function runCommand(command, options) {
   const {ci} = program.opts()
-  return ci ? command.nonInteractive : command.interactive;
+  return ci ? await command.nonInteractive(options) : await command.interactive(options);
 }
```
shikaan, line 12: nit: I think this was good as it was: you're returning whatever the command returns, no need to `await`

shikaan (approved):
Great job 🎉 

There is only one - possible - leftover and a couple of `nit`s (mostly for code hygiene, no blockers)

packages/@contentful--app-scripts/bin/app-scripts
andipaetzold: I always prefer adding await on every level. That way we get a proper stack trace. The outcome when running the code will be the same.

loweisz: I think in the end it doesn't really matter when in both cases `runCommand` will return a promise, or am I missing something ? @andipaetzold what was your intention to use await here?

packages/@contentful--app-scripts/lib/create-app-definition/create-app-definition.js
```diff
@@ -36,18 +36,24 @@ function assertValidArguments(accessToken, appDefinitionSettings) {
   if (!isPlainObject(appDefinitionSettings) || !has(appDefinitionSettings, 'locations')) {
     throwValidationException(
       'AppDefinitionSettings',
-      `Expected plain object with 'location' property, got ${JSON.stringify(appDefinitionSettings, null, 2)}`
-        `Example: ${JSON.stringify({
+      `Expected plain object with 'location' property, got ${JSON.stringify(
+        appDefinitionSettings,
+        null,
+        2
+      )}``Example: ${JSON.stringify(
```
andipaetzold, line 43: Does this syntax with double ` work?

packages/@contentful--app-scripts/lib/upload/create-app-bundle.js
```diff
@@ -64,9 +66,12 @@ async function createAppBundle(settings) {
 
     ${chalk.bold('or by simply running the cli command:')}
 
-      ${chalk.underline(`contentful-app-scripts activate`)}
+      ${chalk.underline(`'npx @contentful/app-scripts activate'`)}
```
andipaetzold, line 69:
```suggestion
      ${chalk.underline('npx @contentful/app-scripts activate')}
```

packages/@contentful--app-scripts/lib/upload/create-app-bundle.js
andipaetzold: The other way round. We use American English in contentful. So: `organization`.
Sorry for being unspecific.

packages/@contentful--app-scripts/lib/create-app-definition/create-app-definition.js
loweisz: I actually haven't changed something here it looks like it came from prettier when I changed something else in the file, so it probably will work when it worked before :D

andipaetzold: Well it was a double ` before, so it didn't change. But e.g. `console.log(`test``test`)` throws an error for me as that's the syntax for template literals: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

loweisz: Yes you are right, I think there was a `,` lost as `throwValidationException` is also accepts three arguments. 
I added it now that the example is the `details` in the function

packages/@contentful--app-scripts/lib/upload/create-app-bundle.js
loweisz: Ahhh I see, okay I changed it now everywhere to `organization`

andipaetzold (approved):

PR #383, opened 2021-04-12
feat(scripts) activate bundle script

loweisz:
Adds the activate script to app-scripts in interactive and in non-interactive (--ci) mode.


packages/@contentful--app-scripts/lib/get-app-info.js
```diff
@@ -0,0 +1,28 @@
...
+const { getManagementToken } = require('./get-management-token');
+
+const getAppInfo = async (options) => {
+  const accessToken = options.token || (await getManagementToken());
+
+  const client = createClient({ accessToken });
+
+  const organization = options.organizationId
+    ? await getOrganizationById(client, options.organizationId)
+    : await selectOrganization(client);
```
shikaan, line 13: Refactoring idea (not really a big deal right now, but worth giving it some thought): what about having a `get$ENTITYId` function which takes care of asking things if needed? Something like

```ts
const getOrganizationId = (organizationId?: string) : string => {
  if (organizationId) {
    return organizationId
  }
  
  // ask all the questions and return the appropriate string
}
```

Then the block above would read:

```ts
const organizationId = await getOrganizationId(options.organizationId);
const organization = getOrganizationById(client, organizationId);
```

This way you wouldn't need to implement two methods for each end every entity. Also, the fetching logic in `getOrganizationId` seems quite general, so we can possibly make that just a single utility function and get rid of a lot of code

packages/@contentful--app-scripts/package.json
```diff
@@ -28,7 +28,7 @@
     "prettier": "prettier **/*.js --write --ignore-path .gitignore",
     "lint": "eslint ./lib",
     "lint:fix": "npm run lint -- --fix",
-    "test": "mocha ./lib/**/*.test.js",
+    "test": "mocha ./lib/**/*.test.js ./lib/*.test.js",
```
shikaan, line 31: Is this needed? The first glob should capture everything 🤔

shikaan (commented):
Looks good to me. Only a couple of minor things to address and a refactoring idea (the latter being not really blocking)

packages/@contentful--app-scripts/lib/get-app-info.js
loweisz: I'm not sure if I understood the approach 100%, but I'm not sure if it would bring more value in that case.
When I understand you correctly, `getOrganizationId` will return the Id in both cases and then fetch it with `getOrganizationById `, to have it unified.

But `selectOrganization` is already returning the organization object what we need. 
So implementing a `getOrganizationId` for both cases and then fetching the organisation in there for all cases would lead to an unnecessary api call when selecting a company and then fetching it again with the id.

Also `getDefinitionById` and `getOrganizationById` look quite similiar, I agree. However there are some slight differences like an additional call in `getDefinitionById` which would make it a bit hard to abstract and might be a bit too much 😬

packages/@contentful--app-scripts/package.json
loweisz: Somehow it wasn't picking it up when being in this directory, did it work for you?

packages/@contentful--app-scripts/README.md
```diff
@@ -70,8 +70,91 @@ provided a Content Management Token (more details [here](https://www.contentful.
...
+| `--definition-id`        | The ID of the app to which to add the bundle                                                                                                   |
+| `--token`                | A personal [access token](https://www.contentful.com/developers/docs/references/content-management-api/#/reference/personal-access-tokens)     |
+
+**Note:** You can also pass all arguments in interactive mode to skip being asked for it.
+
+
+### Activate an AppBundle
+
+Allows you to activate an AppBundle for an AppDefinition.
+When activated the app will serve the newly activated app bundle.
```
andipaetzold, line 125:
```suggestion
When activated the app will serve the newly activated AppBundle.
```

packages/@contentful--app-scripts/README.md
```diff
@@ -70,8 +70,91 @@ provided a Content Management Token (more details [here](https://www.contentful.
...
+### Upload a bundle to an App Definition
+
+Allows you to upload a build directory and create a new AppBundle that is bound to an [AppDefinition](https://www.contentful.com/developers/docs/extensibility/app-framework/app-definition/).
+It runs in interactive or non-interactive mode
+
+**Note:** To make the app serve the bundle you need to [activate](#activate-an-appBundle) it
+
+#### Interactive mode:
+
+In interactive mode, it will ask for all required options
```
andipaetzold, line 88: it = the CLI? I'd be more explicit here.

packages/@contentful--app-scripts/lib/activate/activate-bundle.js
```diff
@@ -0,0 +1,43 @@
+const ora = require('ora');
+const chalk = require('chalk');
+const { throwError } = require('../utils');
+const { createClient } = require('contentful-management');
+
+async function activateBundle({ accessToken, organization, definition, bundleId }) {
+  const activationSpinner = ora('Activating your bundle').start();
+  const client = createClient({ accessToken });
+  const currentOrganization = await client.getOrganization(organization.value);
+  const currentDefinition = await currentOrganization.getAppDefinition(definition.value);
```
andipaetzold, line 10: You can use the plain cma client to reduce the number of requests.

packages/@contentful--app-scripts/lib/activate/activate-bundle.js
```diff
@@ -0,0 +1,43 @@
...
+
+  currentDefinition.bundle = {
+    sys: {
+      id: bundleId,
+      linkType: 'AppBundle',
+      type: 'Link',
+    },
+  };
+  currentDefinition.src = undefined;
```
andipaetzold, line 19: Slightly more correct:
```suggestion
  delete currentDefinition.src;
```

packages/@contentful--app-scripts/lib/activate/activate-bundle.js
```diff
@@ -0,0 +1,43 @@
...
+  try {
+    await currentDefinition.update();
+  } catch (err) {
+    throwError(
+      err,
+      'Something went wrong activating your bundle. Make sure you used the correct definition-id'
```
andipaetzold, line 25:
```suggestion
      'Something went wrong activating your bundle. Make sure you used the correct definition-id.'
```

andipaetzold (changes requested):

packages/@contentful--app-scripts/README.md
```diff
@@ -70,8 +70,91 @@ provided a Content Management Token (more details [here](https://www.contentful.
 
 It only runs in interactive mode.
 
-> **Example** 
-> 
+> **Example**
+>
 > ```shell
 > $ contentful-app-scripts create-app-definition
```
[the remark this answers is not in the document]
loweisz: but according to the package when installed isn't that the right command? https://github.com/contentful/create-contentful-app/blob/c979dea84f8567114228c436718a0f27ec921b55/packages/%40contentful--app-scripts/package.json#L25

I would be okay with installed or npx command, so I can change it to `npx`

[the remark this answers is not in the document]
andipaetzold: That only works if you are installing `contentful-app-scripts` globally.

packages/@contentful--app-scripts/lib/activate/index.js
```diff
@@ -0,0 +1,14 @@
+const { activateBundle } = require('./activate-bundle');
+const { getActivateSettingsArgs } = require('./get-activate-args');
+const { buildBundleActivateSettings } = require('./build-bundle-activate-settings');
+
+module.exports = {
+  async interactive(options) {
+    const settings = await buildBundleActivateSettings(options);
+    activateBundle(settings);
```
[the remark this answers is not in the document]
andipaetzold: @loweisz The `await` is still missing.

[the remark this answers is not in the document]
loweisz: shame on me 😅 . It should be there now 👍

packages/@contentful--app-scripts/lib/activate/activate-bundle.js
```diff
@@ -6,8 +6,12 @@ const { createClient } = require('contentful-management');
...
   const activationSpinner = ora('Activating your bundle').start();
   const client = createClient({ accessToken });
-  const currentOrganization = await client.getOrganization(organization.value);
-  const currentDefinition = await currentOrganization.getAppDefinition(definition.value);
+
+  const definitionUrl = `/organizations/${organization.value}/app_definitions/${definition.value}`;
+
+  const currentDefinition = await client.rawRequest({
+    url: definitionUrl,
+  });
```
andipaetzold, line 14: Not quite what I meant.

Look here for the CMA plain client: https://github.com/contentful/contentful-management.js#alternative-plain-api
You don't need to specify the url yourself.

loweisz: Ahh! wasn't aware of that 👍

packages/@contentful--app-scripts/bin/app-scripts
```diff
@@ -35,6 +35,17 @@ async function runCommand(command, options) {
       await runCommand(upload, options);
     })
 
+  program
+    .command('activate')
+    .description('Activate your AppBundle')
```
shikaan, line 40:
```suggestion
    .description('Mark an AppBundle as "active" for a given AppDefinition')
```

shikaan (approved):
LGTM 🎉

andipaetzold (approved):

PR #386, opened 2021-04-14
feat: [EXT-2674] support node 15

andipaetzold:


Jwhiles:
Currently we're only supporting node 15 - as no node-16 docker image is available for automated testing

shikaan (commented):
:shipit:

PR #387, opened 2021-04-14
chore(caching): Add ability to write env vars to .env file [EXT-2679]

kdamball:
- Writes to `.env` file even if it does't exist
- Adds `.env` to `.gitignore` if possible


packages/@contentful--app-scripts/lib/cache-credential/index.js
```diff
@@ -0,0 +1,89 @@
+const dotenv = require('dotenv');
+const fs = require('fs');
+const { EOL } = require('os');
+const chalk = require('chalk');
+
+
+const fsPromises = fs.promises;
+const fsConstants = fs.constants;
+const envFileName = '.env';
```
jelz, line 9: `const DOTENV_FILE` + do we have some central location for constants?

packages/@contentful--app-scripts/lib/cache-credential/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+// Always export set env vars by default
+dotenv.config();
+
+async function removeOldEnv(envKey) {
+  const envFileData = await fsPromises.readFile(envFileName, {encoding: 'utf-8',});
+  const envVarRegex = new RegExp(`(${EOL}|^)${envKey}=.+(${EOL})?`); // find line with envVar defined
```
jelz, line 16: I wouldn't do regex, just `data.split(newline).filter(line => !line.startsWith(envKey))`

packages/@contentful--app-scripts/lib/cache-credential/index.js
```diff
@@ -0,0 +1,89 @@
...
+    });
+  }
+}
+
+async function addEnvFileToGitIgnore() {
+  try {
+    const gitIgnoreFile = '.gitignore';
+    await fsPromises.access(gitIgnoreFile, fs.constants.F_OK);
+    const gitIgnoreFileData = await fsPromises.readFile(gitIgnoreFile, {encoding: 'utf-8'});
+    const envFileNameRegex = new RegExp(`(^|${EOL})?${envFileName}*?${EOL}?`);
```
jelz, line 30: I would rather use a proper parser (https://www.npmjs.com/package/ignore ?)

packages/@contentful--app-scripts/lib/cache-credential/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+  } catch(err) {
+    console.log(`${
+      chalk.yellow('Warning: could not add .env file to .gitignore. Please don\'t forget to add it manually')
+    }`)
+  }
+}
+
+async function checkFileEnvExists() {
+  let envFileExists = false;
```
jelz, line 47: No need for that, just `return true` in `try` and `return false` outside

packages/@contentful--app-scripts/lib/cache-credential/index.js
andipaetzold, line 9: This will create the file in the current directory and not in the root of the package.
This will cause problems if running `npm run deploy` or similar within the `build` folder.

Same for `.gitignore`

packages/@contentful--app-scripts/lib/cache-credential/index.js
```diff
@@ -0,0 +1,89 @@
+const dotenv = require('dotenv');
+const fs = require('fs');
+const { EOL } = require('os');
+const chalk = require('chalk');
+
+
+const fsPromises = fs.promises;
```
jelz, line 7: I don't think there's a need to drag `Promises` around if we don't use the old API, just name it `fs`

packages/@contentful--app-scripts/lib/cache-credential/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+async function checkFileEnvExists() {
+  let envFileExists = false;
+  try {
+    await fsPromises.access(envFileName, fsConstants.F_OK);
+    envFileExists = true;
+  } catch(_) {
+    envFileExists = false;
+  }
+  return envFileExists;
```
andipaetzold, line 54: It's a little complex, but can be easily simplified:

```suggestion
  try {
    await fsPromises.access(envFileName, fsConstants.F_OK);
    return true;
  } catch(_) {
    return false;
  }
```

packages/@contentful--app-scripts/lib/cache-credential/index.test.js
```diff
@@ -0,0 +1,69 @@
...
+const fs = require('fs');
+const assert = require('assert');
+
+const fsPromises = fs.promises;
+
+describe('Caching environemnt variables', () => {
+  before(async () => {
+    try {
+      await fsPromises.access('.env', fs.constants.F_OK);
+      await fsPromises.copyFile('.env', '.env.temp');
```
jelz, line 11: What happens here? Where does this file come from?

packages/@contentful--app-scripts/lib/cache-credential/index.test.js
```diff
@@ -0,0 +1,69 @@
...
+      await fsPromises.unlink('.env');
+      console.log(`
```
jelz, line 13: I guess these logs need to be removed?

packages/@contentful--app-scripts/lib/cache-credential/index.js
```diff
@@ -0,0 +1,89 @@
...
+async function cacheEnvVars(envObj) {
+  try {
+    let envVars = '';
+    let envFileExists = await checkFileEnvExists();
+
+    for (const [key, val] of Object.entries(envObj)) {
+      if (envFileExists) {
+        await removeOldEnv(key);
+      }
+      envVars += `${EOL}${key}=${val}`
```
andipaetzold, line 66: If the file doesn't exist yet, this makes the file start with an empty line.

packages/@contentful--app-scripts/lib/cache-credential/index.js
```diff
@@ -0,0 +1,89 @@
...
+    if (!envFileNameRegex.test(gitIgnoreFileData)) {
+      await fsPromises.writeFile(gitIgnoreFile, `${EOL}${envFileName}${EOL}`, {
+        encoding: 'utf-8',
+        flag: 'a'
+      });
+    }
+
+  } catch(err) {
+    console.log(`${
+      chalk.yellow('Warning: could not add .env file to .gitignore. Please don\'t forget to add it manually')
```
andipaetzold, line 41: In the upload-bundle PR, we add a period after every message.

```suggestion
      chalk.yellow('Warning: could not add .env file to .gitignore. Please don\'t forget to add it manually.')
```

andipaetzold (changes requested):

packages/@contentful--app-scripts/lib/cache-credential/index.test.js
kdamball: Just in case you have a `.env` lying around. Unlikely but can't be too careful 🤷

packages/@contentful--app-scripts/lib/cache-credential/index.js
kdamball: We do use the old api for `fs.constants`. We can extract both and rename but wanted to avoid confusion considering most projects just call it `fsPromises`

packages/@contentful--app-scripts/lib/cache-credential/index.js
kdamball: Will need to sync on how we use `npm run deploy`. We could use relative to `__dirname` but not sure if this works for `npx`.

packages/@contentful--app-scripts/lib/cache-credential/index.js
kdamball: Deliberately added it as one extra line isn't going to break anything (would also apply when there's EOL in the env file already)

packages/@contentful--app-scripts/lib/cache-credential/index.js
shikaan: I think it's fair to assume the script will be run in the root directory (maybe we can be explicit about that in CCA with a log line or something). 

From `npm-run-script` docs

> Scripts are run from the root of the package folder, regardless of what the current working directory is when npm run is called. If you want your script to use different behavior based on what subdirectory you're in, you can use the INIT_CWD environment variable, which holds the full path you were in when you ran npm run.

So we either are very deliberate with `INIT_CWD` (why should we?) or this looks fine to me

shikaan (commented):
Overall looks good to me, almost ready to go. 

One note here: I would like - feel free to defer - to keep on the top-level `/lib` only the public API of this library to ease discoverability of scripts and encourage contributions. What about moving this script in a sibling `utils` folder? Or anyway an other scaffolding which clearly defines what are the things publicly exposed and what is that we use only internally

packages/@contentful--app-scripts/lib/cache-credential/index.test.js
```diff
@@ -0,0 +1,69 @@
+const { cacheEnvVars } = require('./index');
+const fs = require('fs');
+const assert = require('assert');
+
+const fsPromises = fs.promises;
+
+describe('Caching environemnt variables', () => {
+  before(async () => {
+    try {
+      await fsPromises.access('.env', fs.constants.F_OK);
```
shikaan, line 10: Maybe you can move the whole test run in a temporary folder or just stub the whole `fs` module and replace it with an in-memory representation for the sake of this test. You can use something like `proxyquire` for that (should be already around)

packages/@contentful--app-scripts/lib/create-app-definition/create-app-definition.js
```diff
@@ -64,6 +65,11 @@ async function createAppDefinition(accessToken, appDefinitionSettings = { locati
   ]);
   const selectedOrg = organizations.find((org) => org.value === organizationId);
 
+  await cacheEnvVars({
+    'CONTENTFUL_ORG_ID': organizationId,
+    'CONTENTFUL_ACCESS_TOKEN': accessToken,
+  });
+
```
shikaan, line 72: Are we even reading the variables from the environment once we write them? 🤔 Or is it part of a followup?

packages/@contentful--app-scripts/lib/cache-credential/index.js
andipaetzold: Okay, I assumed running `npm run xx` would be the same as `node ../../lib/index.js`. If that's specified in the docs, this should be fine.

packages/@contentful--app-scripts/lib/cache-credential/index.js
```diff
@@ -0,0 +1,89 @@
...
+
+
+const fsPromises = fs.promises;
+const fsConstants = fs.constants;
+const envFileName = '.env';
+
+// Always export set env vars by default
+dotenv.config();
+
+async function removeOldEnv(envKey) {
```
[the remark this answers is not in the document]
kdamball: This is better for sure. Spent too much time hunting edge cases

packages/@contentful--app-scripts/lib/create-app-definition/create-app-definition.js
kdamball: Lets do a follow up as discussed.

packages/@contentful--app-scripts/package.json
```diff
@@ -28,7 +28,7 @@
     "prettier": "prettier **/*.js --write --ignore-path .gitignore",
     "lint": "eslint ./lib",
     "lint:fix": "npm run lint -- --fix",
-    "test": "mocha ./lib/**/*.test.js ./lib/*.test.js",
+    "test": "mocha ./lib/**/*.test.js ./lib/*.test.js ./utils/**/*.test.js",
     "test:watch": "npm t -- --watch",
     "pre-commit": "lint-staged",
     "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
```
shikaan, line 34: (wrong line, but I cannot comment there) We'd need to add this folder to the `files` list

shikaan (approved):
Looks much cleaner 🎉  (watch out, the CI is red)

kdamball:
@shikaan there's an unrelated test that files because it checks logs. Not sure why that's even happening but checking that one out.

andipaetzold:
@kdamball  What's the purpose of this PR if you do not store the env variables anymore? From what I can see, this PR would only add two dependencies and a js-file that's never used.

kdamball:
@andipaetzold there's a follow up PR for implementing both as we didn't have `upload` ready.

andipaetzold (approved):

PR #391, opened 2021-04-15
Feat: scripts activate bundle by default

loweisz:
Activating the upload automatically and add the option `--skip-activation` in non-interactive mode and also adds a prompt to skip activation in interactive mode. 

The documentation was also changed according to the automatic activation.

### DIFF
--- a/packages/@contentful--app-scripts/lib/upload/build-upload-settings.js
+++ b/packages/@contentful--app-scripts/lib/upload/build-upload-settings.js
@@ -19,14 +19,6 @@ async function buildAppUploadSettings(options) {
       default: '',
     });
   }
-  if (!options.skipActivation) {
-    prompts.push({
-      name: 'skipActivation',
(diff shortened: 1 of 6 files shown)


packages/@contentful--app-scripts/lib/upload/build-upload-settings.js
```diff
@@ -19,13 +19,22 @@ async function buildAppUploadSettings(options) {
       default: '',
     });
   }
+  if (!options.skipActivation) {
```
jelz, line 22: I wouldn't ask for it, only skip if explicitly provided

loweisz: I agree, I wasn't sure about that that's why I added it to be on the safe side. But I will remove it 👍

jelz (approved):

PR #395, opened 2021-04-19
feat(Caching): Using caching on env vars [EXT-2691]

kdamball:
cache Org Id, access token & app def id from create-definition process

## Approach
- First ask and then use env vars as a fallback


packages/@contentful--app-scripts/lib/create-app-definition/get-management-token.js
```diff
@@ -17,6 +21,7 @@ async function getManagementToken() {
     console.log(err.message);
     throw err;
   }
+  const cachedAccessToken = process.env[ACCESS_TOKEN_ENV_KEY];
 
   const { mgmtToken } = await inquirer.prompt([
```
loweisz: Don't we want to skip the prompt when there is a token in the cache?

kdamball: From the discussion and notes, we want to use it as a fallback. @fabe wdyt?

loweisz: I see, it's just that I had the feeling as this is opening the browser to copy the token I'm not sure if it's clear that you can skip that step to use the cached one, but it's just my personal feeling

fabe: Ideally, if there's a token cached we just use that without the prompt. If it fails we ask for it again. I guess that's meant with "fallback" as well?

jelz: I'd also just use it if we've got it stored and retrigger the flow if it's invalid.

If we still ask but use the stored value as the default, then how do we indicate that the default is there? We should never output the value (input is "password", but how does it behave with defaults?)

kdamball: Whenever we cache things - we log loud and clear that we did so. For app definitions & org, we auto-select in interactive mode (user only has to hit enter). It works the same with access token but we dont log anything so, user has to be aware of this behaviour. We can start with cache data as @loweisz & @fabe suggested and resort to original flow if not.

packages/contentful--app-scripts/lib/upload/create-zip-from-directory.js
```diff
@@ -6,9 +6,13 @@ async function createZipFileFromDirectory(path) {
   try {
     const zip = new AdmZip();
     zip.addLocalFolder(path);
-    console.log('----------------------------');
-    console.log(`${chalk.yellow('Done!')} Files from ${chalk.dim(path)} successfully zipped.`);
-    console.log('----------------------------');
+    console.log(`
+  ----------------------------
+
```
shikaan, line 11: very nit: in the logs above we don't have the blank lines between `----`s 😬

shikaan (approved):

packages/contentful--app-scripts/lib/get-management-token.js
```diff
@@ -3,37 +3,65 @@
...
-        }
-
-        return true;
+  const cachedAccessToken = process.env[ACCESS_TOKEN_ENV_KEY];
+  const cachedTokenValid = await checkTokenValidity(cachedAccessToken);
+
+  if (cachedTokenValid) {
+    return cachedAccessToken;
+
+  } else {
```
jelz, line 36: This block is too long, either drop `else` (given `return` above) or extract what's inside to a function.

packages/contentful--app-scripts/lib/upload/create-app-bundle.js
```diff
@@ -30,20 +30,20 @@ async function createAppBundleFromSettings(settings) {
...
+    appUpload.sys.id
+  )}) has been created.`);
   } catch (err) {
     showCreationError('app upload', err.message);
   }
 
   if (!appUpload) return;
 
-  console.log('----------------------------');
+  console.log(`
```
jelz, line 45: I'd rather do EOL or empty `console.log('')

packages/contentful--app-scripts/lib/get-management-token.js
```diff
@@ -3,37 +3,65 @@
...
+const { cacheEnvVars } = require('../utils/cache-credential');
+const { createClient } = require('contentful-management');
+const {
+  ACCESS_TOKEN_ENV_KEY
+} = require('../utils/constants');
+
+const checkTokenValidity = async (accessToken) => {
+  try {
+    const client = createClient({ accessToken });
+    await client.getOrganizations();
```
jelz, line 15: Let's do `getCurrentUser` (it's cheaper)

packages/contentful--app-scripts/lib/get-management-token.js
```diff
@@ -3,37 +3,65 @@
...
+          }
+
+          return true;
+        },
       },
-    },
-  ]);
+    ]);
+
+    await cacheEnvVars({[ACCESS_TOKEN_ENV_KEY]: mgmtToken});
```
loweisz, line 61: Just an idea, wouldn't it make sense to somehow validate the token before caching it? Otherwise it would lead to cache invalid tokens. But it might also be a bit too much

loweisz (approved):

packages/contentful--app-scripts/lib/get-management-token.js
kdamball: We only cache the token pasted in. We can check if it's non-empty but I'm not sure if this is what you mean?

packages/contentful--app-scripts/lib/upload/create-zip-from-directory.js
kdamball: you mean an extra line break was added by mistake? If so, it was to make sure the logs look better 🤷‍♂️ 
Screenshot 2021-04-21 at 12 30 29

packages/contentful--app-scripts/lib/get-management-token.js
loweisz: Yes, but what would happen currently is that we ask for the token and cache it even if it's invalid, then the request fails. 
When the user retries the script the cached token is then picked up and fails at that validation so it will be asked again. 
I'm just wondering if we could save that step to not cache it if it's invalid beforehand. 
But in the end, it probably is not really critical, as the number of requests are the same, so I would be fine with both :)

jelz (approved):

PR #397, opened 2021-04-20
chore: use local templates in ci

shikaan:
## Changes

* remove `@` from template path (to not mess with CRA). This led to remove it from all the packages for consistency
* introduce an environment switch with local resolution and uses the `file:` syntax, as CRA does not seem to cope predictably well with links


package.json
```diff
@@ -21,6 +21,8 @@
     "@commitlint/cli": "^12.0.0",
     "@commitlint/config-conventional": "12.1.1",
     "@octokit/rest": "18.5.2",
+    "@typescript-eslint/eslint-plugin": "^4.22.0",
+    "@typescript-eslint/parser": "^4.22.0",
```
loweisz, line 25: Just out of curiosity, where is this used? 🤔

loweisz (approved):
Really like the approach with the env variable 👍

package.json
shikaan: When we lint the CRA template. I prefer to have dev dependencies hoisted, so that - if we introduce some more ts - it's all on the same version

PR #399, opened 2021-04-20
fix: always use current version for template

loweisz:
Currently, in the CCA script, the template package was hardcoded without a specific version. 

Which leads to, when run as the canary version to not pick up the canary version of the template package. 

This adds the version to the template package as well. All packages should be in sync, so the version of the script should always be the same version as the template.


loweisz:
Will be solved with https://github.com/contentful/create-contentful-app/pull/397

PR #401, opened 2021-04-21
fix: [] use exact version for lerna publish and version

loweisz:
The versions of the internally used local packages should always be strict to avoid some side effects of not being in sync with each other

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,8 +12,8 @@
     "test": "lerna run test",
     "lint": "lerna run lint",
     "link": "lerna link",
-    "version": "lerna version --no-private --conventional-commits --create-release github --yes",
-    "publish": "lerna publish from-git --yes --no-verify-access",
+    "version": "lerna version --no-private --conventional-commits --create-release github --yes --exact",
+    "publish": "lerna publish from-git --yes --no-verify-access --exact",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -12,8 +12,8 @@
     "test": "lerna run test",
     "lint": "lerna run lint",
     "link": "lerna link",
-    "version": "lerna version --no-private --conventional-commits --create-release github --yes",
-    "publish": "lerna publish from-git --yes --no-verify-access",
+    "version": "lerna version --no-private --conventional-commits --create-release github --yes --exact",
+    "publish": "lerna publish from-git --yes --no-verify-access --exact",
```
kdamball, line 16: I might be wrong, but this doesn't seem to be a valid option? https://github.com/lerna/lerna/tree/main/commands/publish

loweisz: I actually think it is. In the docs above the options of `publish`:

 > lerna publish supports all of the options provided by lerna version in addition to the following:
 
 And `--exact` it's defined in `version`
 
 it took me also a while to see that line :D

kdamball: Ah good point - I went straight to the list of options

kdamball (approved):

Issue #403, opened 2021-04-21
Error running npx @contentful/create-contentful-app init


danielstocks:
Hi folks,

I'm trying to create a new Contentful app, using this package, but I keep receiving the following error:

```
➟ npx @contentful/create-contentful-app init my-app
npx: installed 140 in 5.031s
Creating a Contentful app in ~/Projects/my-app.

Creating a new React app in ~/Projects/my-app.

Installing packages. This might take a couple of minutes.
Installing react, react-dom, and react-scripts with @contentful/cra-template-create-contentful-app...

npm ERR! code ETARGET
npm ERR! notarget No matching version found for @contentful/cra-template-create-contentful-app@0.8.1.
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.

npm ERR! A complete log of this run can be found in:
npm ERR!     ~/.npm/_logs/2021-04-21T18_02_59_303Z-debug.log
```

Seems like it's trying to install `@contentful/cra-template-create-contentful-app@0.8.1` which is not published yet.

NPM tells me the [latest available version](https://www.npmjs.com/package/@contentful/cra-template-create-contentful-app) is 0.8.1.alpha.2

loweisz:
Hey Daniel, 

Good catch! There was indeed an issue with the versions in the package. The requested version of the template package was not published. That's why it was failing expecting it to be always in sync. 

I updated all packages with a new version `0.8.2` so the script should work again with latest: 

```
npx create-contentful-app init my-app
```

or 

```
npx @contentful/create-contentful-app init my-app
```

Can you verify that it is working again for you? 
We are also working on a solution, making sure that it won't happen again.

danielstocks:
Hi,

Yeah my short-term solution was just to use an older version (0.8.0) for now:

`npx @contentful/create-contentful-app@0.8.0 create-definition`

I guess the longer term solution is to make sure it doesn't get out of sync again, which you already seem to be working on 👍

PR #404, opened 2021-04-21
fix: issue #403 unable to run @contentful/create-contentful-app init

ShreyasKraman:
**Reason**: package-lock.json was not updated under packages/create-contentful-app folder. 
**Solution**: ran npm install --package-lock-only to fix the issue.


ShreyasKraman:
Fixed in #405

PR #422, opened 2021-05-04
feat: [] adds `open-settings` command to app-scripts

loweisz:
Adds a simple open command to app-scripts. 
The structure of the call is a bit different as there is no sense to have a interactive/non-interactive mode, as this command will only run in one place and never in a CI. 

So there is only one function executed.

### DIFF
--- a/packages/contentful--app-scripts/lib/open/open-settings.js
+++ b/packages/contentful--app-scripts/lib/open/open-settings.js
@@ -35,7 +35,13 @@ async function openSettings(options) {
     throw new Error('No app-definition-id');
   }
 
-  open(`${REDIRECT_URL}&id=${definitionId}`);
+  try {
+    open(`${REDIRECT_URL}&id=${definitionId}`);
+  } catch (err) {
+    console.log(`${chalk.red('Error:')} Failed to open browser`);
+    console.log(err.message);
+    throw err;
+  }
 }
 
 module.exports = {
--- a/packages/contentful--app-scripts/lib/open/open-settings.js
+++ b/packages/contentful--app-scripts/lib/open/open-settings.js
@@ -1,0 +1,29 @@
+const open = require('open');
+const chalk = require('chalk');
+const { APP_DEF_ENV_KEY } = require('../../utils/constants');
+
+const REDIRECT_URL = 'https://app.contentful.com/deeplink?link=app-definition';
+
+async function openSettings(options) {
+  let definitionId;
+  if (options.definitionId) {
+    definitionId = options.definitionId;
+  } else if (process.env[APP_DEF_ENV_KEY]) {
+    definitionId = process.env[APP_DEF_ENV_KEY];
+  } else {
+    console.log(`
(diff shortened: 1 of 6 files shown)


shikaan (commented):
LGTM 🎉 

However, given that `--ci` is a top level flag, I would keep the interactive/non-interactive structure and throw if you run the command in non-interactive mode (as we do for create app definition). Feels more predictable

loweisz:
> LGTM 🎉
> 
> However, given that `--ci` is a top level flag, I would keep the interactive/non-interactive structure and throw if you run the command in non-interactive mode (as we do for create app definition). Feels more predictable

@shikaan I also thought about that, but would you show the prompt to select the definition from the list in interactive mode ? 

Somehow the interactive mode didn't really felt useful to me just to open the settings, as you would need to get the access-token, orgId to have the list of definitions to select. 

Or we make the prompt just a text field to pass in the definition id instead of a list to select from, wdyt?

shikaan:
> > LGTM 🎉
> > However, given that `--ci` is a top level flag, I would keep the interactive/non-interactive structure and throw if you run the command in non-interactive mode (as we do for create app definition). Feels more predictable
> 
> @shikaan I also thought about that, but would you show the prompt to select the definition from the list in interactive mode ?
> 
> Somehow the interactive mode didn't really felt useful to me just to open the settings, as you would need to get the access-token, orgId to have the list of definitions to select.
> 
> Or we make the prompt just a text field to pass in the definition id instead of a list to select from, wdyt?

@loweisz I agree 100% that it's not useful :) I was just speaking about the shape of the code. What I was trying to convey is that we should always keep the command structure predictable (therefore always taking in account the `--ci` flag), but when one of the behaviours - either the interactive or the non-interactive - does not make sense, we throw an exception.

The example I was referring to is this one
https://github.com/contentful/create-contentful-app/blob/7adcb0f3ca538143661be24623d80a9511617d6e/packages/contentful--app-scripts/lib/create-app-definition/index.js#L5-L15

Does it make sense to you?

packages/contentful--app-scripts/lib/open/open-settings.js
```diff
@@ -0,0 +1,44 @@
...
+    console.log(`
+        ${chalk.red('Error:')} There was no app-definition defined.
+
+        Please add it with ${chalk.cyan('--definition-id=<app-definition-id>')}
+        or set the environment variable ${chalk.cyan(`${APP_DEF_ENV_KEY} = <app-definition-id>`)}
+      `);
+    throw new Error('No app-definition-id');
+  }
+
+  open(`${REDIRECT_URL}&id=${definitionId}`);
```
kdamball, line 38: I'd wrap this in a try just in case. We do this in getMgmtToken also

shikaan (approved):

packages/contentful--app-scripts/lib/open/open-settings.js
loweisz: Yes I think that makes sense 👍

kdamball (approved):

PR #428, opened 2021-05-06
feat: [] adds `bundle-cleanup` command to app-scripts

loweisz:
Adds the `bundle-cleanup` to the `app-scripts`. 

Added also the optional `--keep` parameter to adjust the number of bundles to keep instead of hard-coded 50. (50 is still the default when not provided)


packages/contentful--app-scripts/lib/clean-up/clean-up-bundles.js
```diff
@@ -0,0 +1,60 @@
...
+
+${chalk.cyan('Info:')} ${bundlesToDelete.length} bundle${
+    bundlesToDelete.length > 1 ? 's' : ''
+  } will be deleted
+
+  `);
+
+  try {
+    for (const [index, bundle] of bundlesToDelete.entries()) {
+      await deleteBundle(bundle.sys.id, index, bundlesToDelete.length, client, settings);
```
shikaan, line 48: Shall we maybe parallelize and batch this? Not pushing for this specific library, but I was thinking something like [this](https://caolan.github.io/async/v3/docs.html#parallelLimit) to speed up the process and not get 429s

shikaan (commented):
LGTM, maybe we can improve a little on speed here with batching

packages/contentful--app-scripts/utils/constants.js
```diff
@@ -3,9 +3,14 @@ const ACCESS_TOKEN_ENV_KEY = 'CONTENTFUL_ACCESS_TOKEN';
 const ORG_ID_ENV_KEY = 'CONTENTFUL_ORG_ID';
 const APP_DEF_ENV_KEY = 'CONTENTFUL_APP_DEF_ID';
 
+const DEFAULT_BUNDLES_TO_KEEP = 50;
```
kdamball, line 6: Is there a reason we start with such a high number? Basically, is there a risk it will delete their latest active bundle? (I'm not sure about the API validations about deleting an active bundle)

packages/contentful--app-scripts/README.md
```diff
@@ -158,22 +158,65 @@ When passing the `--ci` argument adding all variables as arguments is required
...
+
+In the interactive mode, the CLI will ask for all required options
+
+> **Example**
+>
+> ```shell
+> $ npx --no-install @contentful/app-scripts bundle-cleanup
+> ```
+
+You can also execute this command without the argument if the environment variable (`CONTENTFUL_APP_DEF_ID`) has been set.
```
kdamball, line 197: Isn't ORG_ID also required?

kdamball (approved):
LGTM but minor nit comments

packages/contentful--app-scripts/README.md
loweisz: > I like the idea, but hits may end up being flaky. It'll be more stable if we used fake timers (for example https://sinonjs.org/releases/latest/fake-timers/ )

Yep, you are right 👍  I added fake timers to the timeout

packages/contentful--app-scripts/lib/clean-up/clean-up-bundles.test.js
```diff
@@ -0,0 +1,92 @@
...
+    subject({ ...mockedSettings, keep: 0 });
+    await clock.tickAsync(bundlesFixture.length * 200);
+    deletionCallTimes.forEach((timestamp, index) => {
+      if (index % 2 === 0) {
+        // when the first call in batch should be called approx at the same time as the next one
+        assert.strictEqual(deletionCallTimes[index + 1] - deletionCallTimes[index] < 100, true);
+      } else if (deletionCallTimes[index + 1]) {
+        // when the second call in batch the one after should wait for the first two to finish
+        assert.strictEqual(deletionCallTimes[index + 1] - deletionCallTimes[index] > 100, true);
+      }
```
shikaan, line 89: Maybe with fake timers you don't need the timestamp logic, something like (not to be copy-pasted, more like a suggestion)

```
const clock = useFakeTimers()
clientMock.appBundle.delete = stub().callsFake(({ appBundleId }) => new Promise(r => setTimeout(r, appBundleId === 'slow' ? 500 : 100)))

const bundlesFixture = [
  { sys: { id: 'test-1' } },
  { sys: { id: 'test-2' } },
  { sys: { id: 'test-3' } },
  { sys: { id: 'test-4' } },
  { sys: { id: 'slow' } },
  { sys: { id: 'test-5' } },
  { sys: { id: 'test-6' } },
  { sys: { id: 'test-7' } },
  { sys: { id: 'test-8' } },
];

await subject({ ...mockedSettings, keep: 0 });
await clock.tickAsync(50);
// Here we expect only the first ones to be called
assert.callCount(clientMock.appBundle.delete, 5); // 5 is batch size
await clock.tickAsync(200);
// Here the fast ones have passed, but the slow one is still in-flight
assert.callCount(clientMock.appBundle.delete, 9); // 5 before + 4 new
```

This looks more explicit on what is the behaviour we expect and does not rely on 100ms intervals to assert things. On the other hand, it's a lil more wordy

shikaan (approved):
LGTM I only find the concurrency test hard to read and I have provided a sketch of a suggestion

packages/contentful--app-scripts/lib/clean-up/clean-up-bundles.test.js
loweisz: I actually like this approach 👍  The fake timers make it easier to test it. 
I just splitted it into two tests to have a better overview

PR #475, opened 2021-06-01
feat: [EXT-2678] add validation for build folder

Jwhiles:
The upload script will now throw if there is no index.html file, and will warn if there are relative paths.

### DIFF
--- a/packages/contentful--app-scripts/lib/upload/validate-bundle.js
+++ b/packages/contentful--app-scripts/lib/upload/validate-bundle.js
@@ -26,7 +26,7 @@ const validateBundle = path => {
     console.warn(
       `${chalk.red(
         'Warning:'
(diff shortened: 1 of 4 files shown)


packages/contentful--app-scripts/lib/upload/validate-bundle.js
```diff
@@ -0,0 +1,37 @@
...
+
+  const entryFile = fs.readFileSync(Path.join(buildFolder, entry), { encoding: 'utf8' });
+
+  if (fileContainsAbsolutePath(entryFile)) {
+    console.log('----------------------------');
+    console.warn(
+      `${chalk.red(
+        'Warning:'
+      )} This bundle uses absolute paths. Please use relative paths instead for correct rendering`
+    );
```
loweisz, line 30: Maybe we could also add the same link as we did in the web app?

loweisz (approved):

PR #502, opened 2021-06-07
chore: fix dependabot/lerna config

andipaetzold:
- single dependabot root
  This should be the correct way to use dependabot according to https://github.com/github/docs/discussions/596#discussioncomment-105956
- remove invalid arguments and don't force version
  - `--exact` and `--force-publish` don't exist on `lerna publish`
  - `--force-publish` doesn't exist on `lerna version` and might (not 100% sure) the reason why always a new version is released
- use relative paths for local packages
  Dependabot doesn't bump other packages. By using relative paths, we avoid having to bump the version in all dependants
- downgrade package-lock to v1
  50% of the packages were accidentally bumped to npm 6.
- only run `npm ci` && `npm run bootstrap` in CI


packages/create-contentful-app/package.json
```diff
@@ -25,7 +25,7 @@
     "create-contentful-app": "index.js"
   },
   "dependencies": {
-    "@contentful/create-contentful-app": "0.9.10"
+    "@contentful/create-contentful-app": "file:../contentful--create-contentful-app"
```
shikaan, line 28: Does this work with npm (as opposed to yarn)? You might get 404s

andipaetzold: Still testing. What I know is that dependabot doesn't bump the version of other monorepo packages.

andipaetzold: lerna is using `file:` within their own repo (https://github.com/lerna/lerna) and use npm

andipaetzold:
To solve the issue of dev dep updates bumping the version, we might want to move these deps to the root package. That way we also reduce the total size of `node_module`s.

shikaan (approved):

PR #561, opened 2021-07-08
feat: [EXT-2477] Use react function component instead of class

colomolo:


packages/contentful--cra-template-create-contentful-app/template/src/components/ConfigScreen.tsx
```diff
@@ -9,59 +9,58 @@ interface ConfigProps {
...
-          <Paragraph>Welcome to your contentful app. This is your config page.</Paragraph>
-        </Form>
-      </Workbench>
-    );
-  }
+  useEffect(() => {
+    // `onConfigure` allows to configure a callback to be
+    // invoked when a user attempts to install the app or update
+    // its configuration.
+    props.sdk.app.onConfigure(() => onConfigure());
```
shikaan, line 37: Should the next reviewer in line also trip on this: `onConfigure` accepts only one subscription. Even when we do not unsubscribe, this won't leak

shikaan (approved):
Looks great to me 🎉

PR #563, opened 2021-07-09
chore: run npm install before checking outdated deps

andipaetzold:
If you run `npm outdated` before `npm install`, only `dependencies` are displayed as outdated. Only if you run `npm install` also `devDependencies` are listed as outdated.
I didn't find any documentation or a cli flag for this, so I added `npm install` to the CI when updating the `template.json`.

In our case `@contentful/app-scripts` is a few versions behind.


Jwhiles (approved):
You learn something new everyday

PR #571, opened 2021-07-14
feat: [] add cma adapter example to readme

loweisz:
This adds the new cma adapter functionality as an example to the template

For that, I also added `contentful-management` as a dependency.

**Thoughts:**
Maybe we could instead add in every location component an example, but I didn't want to bloat it too much.
Even though I'm really a fan of how the init function in `index.ts` is growing with that and people might oversee what this is and just delete it.


packages/contentful--cra-template-create-contentful-app/template/src/index.tsx
```diff
@@ -32,8 +32,8 @@ if (process.env.NODE_ENV === 'development' && window.self === window.top) {
   render(<LocalhostWarning />, root);
 } else {
   init((sdk) => {
-    // How to create an instance of the the contentful-management-api client
-    // Can be removed together with the contentful-management package if it is not needed
+    // You can create an instance of the the contentful-management client.
+    // This and the contentful-management package can be removed if its not used
     // Visit https://www.contentful.com/developers/docs/extensibility/app-framework/sdk/#using-the-contentful-management-api
```
andipaetzold, line 37: The url is not correct anymore

packages/contentful--cra-template-create-contentful-app/template/src/index.tsx
```diff
@@ -32,8 +32,8 @@ if (process.env.NODE_ENV === 'development' && window.self === window.top) {
   render(<LocalhostWarning />, root);
 } else {
   init((sdk) => {
-    // How to create an instance of the the contentful-management-api client
-    // Can be removed together with the contentful-management package if it is not needed
+    // You can create an instance of the the contentful-management client.
```
andipaetzold, line 35: Remove this period `.` everywhere or remove it here.

andipaetzold (changes requested):
Looks good. Just two small comments

packages/contentful--cra-template-create-contentful-app/template/src/index.tsx
```diff
@@ -31,6 +32,24 @@ if (process.env.NODE_ENV === 'development' && window.self === window.top) {
...
+      {
+        type: 'plain',
+        defaults: {
+          environmentId: sdk.ids.environment,
+          spaceId: sdk.ids.space,
+        },
+      }
+    );
+
+    // Use the plain or nested client without any additional configuration
```
Jwhiles, line 50: This feels a bit disconnected to me, and doesn't really explain itself.

I think change it to something like
```suggestion
    // You can use the cma client created above without any additional configuration
    // for example you can call cma.locale.getMany({}) to get a list of locales
```

Or just remove this entirely, as we have the link above explaining what the CMA client is for.

fabe:
Just my 2c: I'd generally want us to include less bloat in here. Especially in the init function, should you even do any actions against the CMA? Unless you pass down the client to the locations I don't really see a use case we'd want to promote here. I'd probably prefer adding this snippet to the readme or documentation instead.

loweisz:
> Just my 2c: I'd generally want us to include less bloat in here. Especially in the init function, should you even do any actions against the CMA? Unless you pass down the client to the locations I don't really see a use case we'd want to promote here. I'd probably prefer adding this snippet to the readme or documentation instead.

I actually agree with that. I'm not a fan of dead code in the template either. I will move it to the Readme👍

packages/contentful--cra-template-create-contentful-app/template/README.md
```diff
@@ -45,6 +45,33 @@ To make your app look and feel like Contentful use the following libraries:
...
+  }
+);
+
+cma.locale.getMany({})
+```
+
+Visit [the official docs](https://www.contentful.com/developers/docs/extensibility/app-framework/sdk/#using-the-contentful-management-api)
+to find out more.
+
+_Note:_ Feel free to uninstall the `contentful-management` package if you are not using the client.
```
andipaetzold, line 72: Should we even install it?

packages/contentful--cra-template-create-contentful-app/template/README.md
```diff
@@ -45,6 +45,33 @@ To make your app look and feel like Contentful use the following libraries:
...
+      environmentId: sdk.ids.environment,
+      spaceId: sdk.ids.space,
+    },
+  }
+);
+
+cma.locale.getMany({})
+```
+
+Visit [the official docs](https://www.contentful.com/developers/docs/extensibility/app-framework/sdk/#using-the-contentful-management-api)
```
andipaetzold, line 69: I assume the fragment is still broken and needs to be updated.

packages/contentful--cra-template-create-contentful-app/template/README.md
```diff
@@ -45,6 +45,33 @@ To make your app look and feel like Contentful use the following libraries:
 - [Forma 36](https://f36.contentful.com/) – Contentful's design system
 - [Contentful Field Editors](https://www.contentful.com/developers/docs/extensibility/field-editors/) – Contentful's field editor React components
 
+## Using the contentful-management SDK
+
+With the sdk and the contentful-management package installed,
+you can also create an instance of the contentful-management client using the `cmaAdapter`, which is part of the sdk, without additional configuration.
```
andipaetzold, line 51: "without additional configuration" is followed by some code with additional configuration. `{type: 'plain'...`

packages/contentful--cra-template-create-contentful-app/template/README.md
```diff
@@ -45,6 +45,33 @@ To make your app look and feel like Contentful use the following libraries:
...
+
+
+```
```
andipaetzold, line 54:
```suggestion
```javascript
```

packages/contentful--cra-template-create-contentful-app/template/README.md
andipaetzold: There are no unofficial docs, right?

```suggestion
Visit [the docs](https://www.contentful.com/developers/docs/extensibility/app-framework/sdk/#using-the-contentful-management-api)
```

andipaetzold (commented):
Just some more nit picking 😁 

We should write `contentful-management` instead of contentful-management

packages/contentful--cra-template-create-contentful-app/template/README.md
loweisz: I think having it installed doesn't hurt as it is installed with the `peerDependency` of `app-sdk` anyway and maybe we can push it with that a bit more to be used.
But would also be fine to remove it. @fabe wdyt?

packages/contentful--cra-template-create-contentful-app/template/README.md
loweisz: changed it to without the access token, as this is what I meant with it

packages/contentful--cra-template-create-contentful-app/template/README.md
andipaetzold: Peer Dependencies are not automatically installed.

But maybe we should make it an actual dependency of App SDK as we are relying on the CMA types.

loweisz: Nevertheless, it will be installed from somewhere when you check the node modules, even if it is not defined as a dependency.

packages/contentful--cra-template-create-contentful-app/template/README.md
```diff
@@ -45,6 +45,33 @@ To make your app look and feel like Contentful use the following libraries:
 - [Forma 36](https://f36.contentful.com/) – Contentful's design system
 - [Contentful Field Editors](https://www.contentful.com/developers/docs/extensibility/field-editors/) – Contentful's field editor React components
 
+## Using the contentful-management SDK
```
shikaan, line 48: What about we do not introduce cma here and we add

```
npm i @contentful/contentful-management@latest @contentful/app-sdk@canary
```

in the instructions? The setup we provide won't work with app-sdk v3 and I wouldn't give the out-of-the-box option until app-sdk is stable.

andipaetzold (approved):

packages/contentful--cra-template-create-contentful-app/template/README.md
```diff
@@ -45,6 +45,33 @@ To make your app look and feel like Contentful use the following libraries:
...
+
+With the sdk and the contentful-management package installed,
+you can also create an instance of the contentful-management client using the `cmaAdapter`, which is part of the sdk, without passing the access token.
+
+
+```js
+const cma = createClient(
```
fabe, line 55: Add `import` for `createClient`

packages/contentful--cra-template-create-contentful-app/template/README.md
```diff
@@ -45,6 +45,33 @@ To make your app look and feel like Contentful use the following libraries:
 - [Forma 36](https://f36.contentful.com/) – Contentful's design system
 - [Contentful Field Editors](https://www.contentful.com/developers/docs/extensibility/field-editors/) – Contentful's field editor React components
 
+## Using the contentful-management SDK
+
+With the sdk and the contentful-management package installed,
+you can also create an instance of the contentful-management client using the `cmaAdapter`, which is part of the sdk, without passing the access token.
```
fabe, line 51:
```suggestion
you can also create an instance of the `contentful-management` client using the `cmaAdapter`, which is part of the SDK, without passing the access token.
```

packages/contentful--cra-template-create-contentful-app/template/README.md
```diff
@@ -45,6 +45,33 @@ To make your app look and feel like Contentful use the following libraries:
 - [Forma 36](https://f36.contentful.com/) – Contentful's design system
 - [Contentful Field Editors](https://www.contentful.com/developers/docs/extensibility/field-editors/) – Contentful's field editor React components
 
+## Using the contentful-management SDK
+
+With the sdk and the contentful-management package installed,
```
fabe, line 50:
```suggestion
With the SDK and the `contentful-management` package installed,
```

packages/contentful--cra-template-create-contentful-app/template/README.md
```diff
@@ -45,6 +45,33 @@ To make your app look and feel like Contentful use the following libraries:
...
+      environmentId: sdk.ids.environment,
+      spaceId: sdk.ids.space,
+    },
+  }
+);
+
+cma.locale.getMany({})
+```
+
+Visit [the docs](https://www.contentful.com/developers/docs/extensibility/app-framework/sdk/#using-the-contentful-management-library)
```
fabe, line 69:
```suggestion
Visit the [`contentful-management` documentation](https://www.contentful.com/developers/docs/extensibility/app-framework/sdk/#using-the-contentful-management-library)
```

packages/contentful--cra-template-create-contentful-app/template/README.md
fabe:
```suggestion
## Using the `contentful-management` SDK
```

fabe (approved):

PR #716, opened 2021-11-05
fix: template dependencies

loweisz:
- I moved the types and testing dependencies to `devDependencies`
- removed `@contentful/field-editor-test-utils` and `@contentful/field-editor-single-line` as they were not used. 
- Removed `@testing-library/user-event` as it is not used
- Removed emotion from the code


packages/contentful--cra-template-create-contentful-app/template.json
```diff
@@ -1,25 +1,27 @@
...
+      "@contentful/forma-36-fcss": "^0.3.4",
+      "@contentful/forma-36-react-components": "^3.100.1",
+      "@contentful/forma-36-tokens": "^0.11.1",
+      "contentful-management": "^7.45.0",
+      "cross-env": "^7.0.3",
+      "emotion": "^10.0.17",
+      "react": "^17.0.2",
+      "react-dom": "^17.0.2",
+      "react-scripts": "4.0.3",
+      "typescript": "^4.4.4"
```
andipaetzold, line 14: `typescript` and `cross-env` should also be a dev dependency

packages/contentful--cra-template-create-contentful-app/template.json
```diff
@@ -1,25 +1,27 @@
...
+      "react": "^17.0.2",
+      "react-dom": "^17.0.2",
+      "react-scripts": "4.0.3"
     },
     "devDependencies": {
-      "@contentful/app-scripts": "^0.13.0"
+      "@contentful/app-scripts": "^0.13.37",
+      "@testing-library/jest-dom": "^5.15.0",
+      "@testing-library/react": "^12.1.2",
+      "@testing-library/user-event": "^13.5.0",
```
andipaetzold, line 18: We can remove `@testing-library/user-event`

andipaetzold (commented):
> I manually installed emotion as it is used in the config location but only cam through one of the above packages, so I made this dependency a bit more explicit.

It's used in the code once for a simple margin... feels wrong to use emotion here...

Jwhiles (approved):

PR #721, opened 2021-11-09
feat: update template structure []

Chaoste:
After working on the tasks-app, we found a few improvements that we would like to have in the general template.

Commits:
- feat: change template to have a deeper structure
- feat: configure default testId in template for testing-library
- feat(template): update imports and improve types
- feat(template): enhance gitignore

### DIFF
--- a/packages/contentful--cra-template-create-contentful-app/template/gitignore
+++ b/packages/contentful--cra-template-create-contentful-app/template/gitignore
@@ -10,6 +10,7 @@
 
 # production
 /build
+/dist
 
 # misc
 .DS_Store
--- a/packages/contentful--cra-template-create-contentful-app/template/gitignore
+++ b/packages/contentful--cra-template-create-contentful-app/template/gitignore
@@ -10,7 +10,6 @@
 
 # production
 /build
-/dist
 
 # misc
 .DS_Store
--- a/packages/contentful--cra-template-create-contentful-app/template/src/setupTests.ts
+++ b/packages/contentful--cra-template-create-contentful-app/template/src/setupTests.ts
@@ -3,3 +3,8 @@
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
 import '@testing-library/jest-dom';
+import { configure } from '@testing-library/react';
+
+configure({
+  testIdAttribute: 'data-test-id',
+});
--- a/packages/contentful--cra-template-create-contentful-app/template/src/components/ConfigScreen.spec.tsx
+++ b/packages/contentful--cra-template-create-contentful-app/template/src/components/ConfigScreen.spec.tsx
@@ -1,11 +1,11 @@
 import React from 'react';
 import ConfigScreen from './ConfigScreen';
 import { render } from '@testing-library/react';
-import { mockSdk } from '../../test/mocks/mockSdk';
+import { mockCma, mockSdk } from '../../test/mocks';
 
 describe('Config Screen component', () => {
   it('Component text exists', async () => {
-    const { getByText } = render(<ConfigScreen sdk={mockSdk} />);
+    const { getByText } = render(<ConfigScreen cma={mockCma} sdk={mockSdk} />);
 
     // simulate the user clicking the install button
     await mockSdk.app.onConfigure.mock.calls[0][0]();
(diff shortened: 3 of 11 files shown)


Jwhiles:
Do you think that making the folder structure more deeply nested really helps developer onboarding? To me it seems like we're making code discover marginally harder with little obvious benefit.

Chaoste:
> Do you think that making the folder structure more deeply nested really helps developer onboarding? To me it seems like we're making code discover marginally harder with little obvious benefit.

I would say it really depends whether you want to use this template for a small use case with a few lines of code or if it's meant to be used for massive product development (like tasks-app or workflows-marketplace-app). In both cases, we're bootstrapping with this repo but then rewriting a lot of code and config. Instead, I would love to see better coding practices (like max. 10 files per folder) in this template as it's also used by our customers.

Chaoste:
@Jwhiles Hey John, let me know if you any further questions on this, so I can unblock it (after fixing the tests) :)

Jwhiles:
Hey @Chaoste - I think some members of the team @andipaetzold and @loweisz are looking into making some more far reaching changes to CCA, and that making these changes to the structure of the output are just adding extra churn to something that will be changing more dramatically anyway. 

I think if you're happy to remove the structural changes, we'd be happy to add the test-id, gitignore, and type improvements.

Chaoste:
> Hey @Chaoste - I think some members of the team @andipaetzold and @loweisz are looking into making some more far reaching changes to CCA, and that making these changes to the structure of the output are just adding extra churn to something that will be changing more dramatically anyway.
> 
> I think if you're happy to remove the structural changes, we'd be happy to add the test-id, gitignore, and type improvements.

Thanks for the feedback. I removed the folder structure and added a fix to the tests to remove TS errors(cma prop was missing)

Lelith (approved):
Looks good to me!

packages/contentful--cra-template-create-contentful-app/template/gitignore
```diff
@@ -10,6 +10,7 @@
 
 # production
 /build
+/dist
```
andipaetzold, line 13: Why? Create React App always builds to `build`.

Chaoste: Mmmh, it's not necessary for every fork of the project but in general `dist` is an output folder. We "needed" it for the tasks-app when we want to have the deploy step in [this PR](https://github.com/contentful/tasks-app/pull/33/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R61).

Chaoste:
```suggestion
```

andipaetzold (approved):
This doesn't conflict with the planned changes to the template. 👍

packages/contentful--cra-template-create-contentful-app/template/src/setupTests.ts
```diff
@@ -3,3 +3,8 @@
 // expect(element).toHaveTextContent(/react/i)
 // learn more: https://github.com/testing-library/jest-dom
 import '@testing-library/jest-dom';
+import { configure } from '@testing-library/react';
+
+configure({
+  testIdAttribute: 'data-test-id',
```
andipaetzold, line 9: Sorry for coming back to this PR even though it is already merged.

I thought about the CCA yesterday and I don't think that we should customize the default jest /testing-library behavior here and just leave the defaults. Yes, Contentful uses `data-test-id` for setting test ids on elements. But as this package is used by many users outside of Contentful, a change to a default is more likely to cause confusion and problems.

Lelith: it confused me when joining contentful, questioning if i forgot how to use getByTestId selector 😅

andipaetzold: wdyt @Chaoste ?

Chaoste: Mmmh, I don't have a strong opinion. I guess my intention is to align our public repositories, namely "forma-36" and this one. For forma-36, we somehow "force" people to do this (if they need to reference our f36 components in their tests).

andipaetzold: I wasn't aware of how F36 assigns test-ids. So I think this is fine as we at least stay consistent with other Contentful repos.
Still unfortunate to change a default here...

PR #754, opened 2021-12-10
feat: bump template to use forma v4

massao:


gui-santos (approved):

bgutsol (approved):

packages/contentful--cra-template-create-contentful-app/template.json
```diff
@@ -2,9 +2,8 @@
   "package": {
     "dependencies": {
       "@contentful/app-sdk": "^4.3.0",
-      "@contentful/forma-36-fcss": "^0.3.2",
-      "@contentful/forma-36-react-components": "^3.90.7",
-      "@contentful/forma-36-tokens": "^0.11.0",
+      "@contentful/f36-components": "latest",
+      "@contentful/f36-tokens": "latest",
```
andipaetzold, line 6: The `package.json` should always have specific version numbers. We don't want to create new apps using `latest`.

andipaetzold (changes requested):
This PR must be merged with a commit message other than `chore:`. Using `chore`, no new version will be released.

packages/contentful--cra-template-create-contentful-app/template/src/components/ConfigScreen.tsx
```diff
@@ -1,8 +1,8 @@
 import React, { useCallback, useState, useEffect } from 'react';
 import { AppExtensionSDK } from '@contentful/app-sdk';
 import { PlainClientAPI } from 'contentful-management';
-import { Heading, Form, Workbench, Paragraph } from '@contentful/forma-36-react-components';
-import styles from './ConfigScreen.module.css';
+import { Heading, Form, Paragraph, Flex } from '@contentful/f36-components';
+import { css } from 'emotion';
```
andipaetzold, line 5: `emotion` is not part of the template anymore. Please keep using the CSS module

andipaetzold (changes requested):

packages/contentful--cra-template-create-contentful-app/template/src/components/ConfigScreen.tsx
suevalov: @andipaetzold `emotion` is a recommended way to style Forma v4 components and I think we need to re-evaluate the decision

andipaetzold: `emotion` was only used in this file and we saw that as an opportunity to get rid of a dependency and make the bundle smaller. If we want to reintroduce `emotion` with this PR, we'd need to add it back to `template.json`.

massao: I believe that we want to add it back as it's the recommended library to be used with forma v4 components, right @burakukula and @suevalov ??

suevalov: @massao yes, I'd do that

andipaetzold (changes requested):
Please remove the `yarn.lock` again

suevalov (approved):

andipaetzold (approved):

PR #759, opened 2021-12-15
fix: bump create-react-app and react-scripts

loweisz:
Bump create-react-app and react-scripts to v5


kdamball (approved):

andipaetzold (approved):
LGTM (if CI passes)

Should be a `fix:` when being merged

PR #769, opened 2021-12-21
fix(tests): ignore typecheck for test which breaks dev-build

ruderngespra:
Adding mock for `Fields.spec.ts` as it currently breaks when running `npm start` breaks due to an enforced type check.

image


kdamball (commented):
Can we mock the sdk like other files?

ruderngespra:
> Can we mock the sdk like other files?

@kdamball Very good point, thanks, I overlooked that we had those mocks ready for use.

kdamball (approved):

PR #808, opened 2022-01-20
fix: remove css file

andipaetzold:
We removed the import in https://github.com/contentful/create-contentful-app/pull/754 but not the file itself


kdamball (approved):
First time I'm seeing this file 👀

andipaetzold:
> First time I'm seeing this file 👀

@kdamball  I added it when I thought I was clever and remove emotion from the code base 😬

PR #845, opened 2022-02-07
feat: replace create-react-app with degit [EXT-3404]

andipaetzold:
This PR removes create-react-app as dependency and clones the template using `degit`.

https://github.com/contentful/apps/pull/735 must be merged first, so the path can be adjusted

With this PR I switched to ES modules to make the imports prettier.

There will be a follow-up cleanup PR that adjusts CI and removes the old template: https://github.com/contentful/create-contentful-app/pull/846

## TODO

- [x] Adjust the name in `package.json`


packages/contentful--create-contentful-app/package.json
```diff
@@ -36,10 +37,8 @@
     "@contentful/cra-template-create-contentful-app": "file:../contentful--cra-template-create-contentful-app",
```
shikaan, line 37: we might need to remove this

packages/contentful--create-contentful-app/lib/index.js
```diff
@@ -2,54 +2,56 @@
...
+    if (e.code === 'DEST_NOT_EMPTY') {
+      message = 'destination directory is not empty';
+    }
+    throw new Error(message);
+  }
 }
 
-function initProject() {
+async function initProject() {
+  const appFolder = process.argv[3];
```
shikaan, line 52: should we maybe validate if this is a correct path-like string?

packages/contentful--create-contentful-app/lib/index.js
```diff
@@ -60,29 +62,25 @@ function initProject() {
...
-        onSuccess(appFolder);
-      } else {
-        process.exit(exitCode);
-      }
-    });
+    rmIfExists(resolve(fullAppFolder, 'package-lock.json'));
+    rmIfExists(resolve(fullAppFolder, 'yarn.lock'));
+    updatePackageName(fullAppFolder);
+
+    await exec('npm', ['install'], { cwd: fullAppFolder });
```
shikaan, line 76: nit: maybe we can check if npm version is >=6 as we require in the package json? I am not sure if we get any warnings or anything when executed as child process

shikaan (commented):
LGTM, few minor comments

packages/contentful--create-contentful-app/package.json
andipaetzold: Yes forgot about that and did it in the follow-up PR: https://github.com/contentful/create-contentful-app/pull/846/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L52

packages/contentful--create-contentful-app/lib/index.js
andipaetzold: The `engines` property in package.json of `create-contentful-app` should show a warning like this:

```
npm WARN EBADENGINE Unsupported engine {
npm WARN EBADENGINE   package: 'create-contentful-app@1.0.0',
npm WARN EBADENGINE   required: { node: '>=16.0.0' },
npm WARN EBADENGINE   current: { node: 'v14.15.0', npm: '7.5.3' }
npm WARN EBADENGINE }
```

packages/contentful--create-contentful-app/lib/index.js
andipaetzold: What is a path-like string?

Should `path.resolve(...)` handle all these cases and just return a proper absolute path?

packages/contentful--create-contentful-app/lib/index.js
shikaan: Yeah, I was more thinking about an explicit check for wen we incorporate yarn, for example. We can handle that then though

andipaetzold: Yes, we might want to check whether `yarn` is installed. And even when using `yarn`, we probably still have a minimum npm version.
Let's do this in a future PR.

packages/contentful--create-contentful-app/lib/index.js
shikaan: I don't think we get any validation there

**Node REPL**
```
path.resolve(process.cwd(), '>>><>?||')
'$ABSOLUTE_PATH/>>><>?||'
```

But on Windows, for example, `\ / : * ? " < > |` are invalid chars for filenames. Maybe it's handled in the windows version of Node (and therefore `path`)

shikaan (approved):

PR #852, opened 2022-02-10
feat: Add flags support [EXT-2743]

colomolo:


packages/contentful--create-contentful-app/lib/index.js
```diff
@@ -49,20 +48,20 @@ async function cloneTemplate(name, destination) {
...
       throw new Error(
         `Cannot create an app named "${appName}". Please choose a different name for your app.`
       );
     }
 
+    if (options.npm && options.yarn) {
+      throw new Error(
+        `Please provide either ${chalk.bold('--yarn')} or ${chalk.bold('--npm')} flag, but not both`
+      );
+    }
```
andipaetzold, line 63: When both flags are set, we want to use npm. That's the same as we do it when `--javascript `and `--typescript` is set.

Right? @fabe

packages/contentful--create-contentful-app/lib/index.js
```diff
@@ -86,44 +91,27 @@ async function initProject() {
...
-  }
-
-  switch (command) {
-    case 'init':
-      await initProject();
-      break;
-
-    case 'create-definition':
+  program
+    .command('init', { isDefault: true, hidden: true })
```
andipaetzold, line 95: Why `hidden`?

andipaetzold (changes requested):
Looks good. Just two comments.

packages/contentful--create-contentful-app/lib/index.js
```diff
@@ -94,7 +94,7 @@ async function initProject(appName, options) {
   program
     .command('init', { isDefault: true, hidden: true })
     .description('Bootstrap your app inside a new folder ‘app-name’')
-    .argument('[app-name]', 'App name')
+    .argument('[app-name]', 'App name', 'my-app')
```
andipaetzold, line 97: Didn't we say we want to start interactive mode when it's missing?

packages/contentful--create-contentful-app/lib/index.js
fabe: Yes, let's default npm in that case!

fabe: Maybe we can log a warning like `Provided both --yarn and --npm, using `--npm.`?

packages/contentful--create-contentful-app/lib/index.js
```diff
@@ -50,39 +48,38 @@ async function cloneTemplate(name, destination) {
...
-
     if (!validateAppName(appName).validForNewPackages) {
       throw new Error(
         `Cannot create an app named "${appName}". Please choose a different name for your app.`
       );
     }
 
+    if (options.npm && options.yarn) {
+      throw new Error(
+        `Please provide either ${chalk.bold('--yarn')} or ${chalk.bold('--npm')} flag, but not both`
```
fabe, line 61:
```suggestion
        `Please provide either ${chalk.bold('--yarn')} or ${chalk.bold('--npm')}.`
```
Maybe less verbose?

packages/contentful--create-contentful-app/lib/index.js
colomolo: And also add the same warning to js vs ts conflict? WDYT?

packages/contentful--create-contentful-app/lib/index.js
```diff
@@ -94,30 +91,30 @@ async function initProject() {
...
+    .action(async () => {
       await createAppDefinition.interactive();
-      break;
+    });
 
-    case 'help':
-      printHelpText(mainCommand, localCommand);
-      break;
+  program
+    .name(chalk.cyan('@contentful/create-contentful-app'))
```
fabe, line 115:
```suggestion
    .name(chalk.cyan('create-contentful-app'))
```
Should we get rid of the scope here as well @andipaetzold ?

packages/contentful--create-contentful-app/lib/index.js
colomolo: This will be done in the follow-up PR

packages/contentful--create-contentful-app/lib/index.js
andipaetzold: Yes. I wanted to create a ticket to have a proper cleanup (including docs), so we are more consistent.

packages/contentful--create-contentful-app/lib/utils.js
```diff
@@ -15,25 +15,19 @@ export function exec(command, args, options) {
...
-}
-
 export function rmIfExists(path) {
   if (existsSync(path)) {
     rmSync(path);
   }
 }
+
+export function detectManager() {
+  switch (basename(process.env.npm_execpath || '')) {
```
kdamball, line 25: Is this something that is always available or do we need to set it?

colomolo: Probably it is always available, but if not we fallback to npm.

kdamball: `node -pe "process.env.npm_execpath"` is `undefined` for me. Can we confirm that this is something that's commonly available?

colomolo: it is available only if you use npm or yarn.

kdamball: Ah - good point. I see it also if I run it thru a script

kdamball (approved):

packages/contentful--create-contentful-app/lib/index.js
```diff
@@ -8,12 +8,10 @@ import degit from 'degit';
 import { readFileSync, writeFileSync } from 'fs';
 import { basename, resolve } from 'path';
 import tildify from 'tildify';
+import { exec, rmIfExists, detectManager } from './utils.js';
```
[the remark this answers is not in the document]
colomolo: You mean help? Yes, it is autogenerated.

PR #853, opened 2022-02-10
feat(JSTempalte): Support JS templates [EXT-3441]

kdamball:
Support template flag for typescript(`--typescript`/`-ts`)  & JS(`--javascript`/`-js`)

Usage: `@contentful/create-contentful-app init my-contentful-app --javascript`

### DIFF
--- a/packages/contentful--create-contentful-app/lib/index.js
+++ b/packages/contentful--create-contentful-app/lib/index.js
@@ -8,9 +8,10 @@ import degit from 'degit';
 import { readFileSync, writeFileSync } from 'fs';
 import { basename, resolve } from 'path';
 import tildify from 'tildify';
-import { exec, rmIfExists } from './utils.js';
 import os from 'os';
 import validateAppName from 'validate-npm-package-name';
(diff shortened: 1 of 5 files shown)


packages/contentful--create-contentful-app/lib/index.js
```diff
@@ -67,7 +68,16 @@ async function initProject() {
 
     console.log(`Creating a Contentful app in ${chalk.bold(tildify(fullAppFolder))}.`);
 
-    await cloneTemplate('typescript', fullAppFolder);
+    program
+      .option('--javascript, -js')
+      .option('--typescript, -ts')
+      .action(async (opts) => {
+        console.log()
```
andipaetzold, line 75: 🤔  why is that here?

andipaetzold (approved):
Seems to work fine 👍

andipaetzold (approved):

andipaetzold:
I think you need to run `npm i` on root level and then also tests should be green.

kdamball:
Lerna strikes again

PR #868, opened 2022-02-14
feat: EXT-3461 support custom templates

shikaan:
This touches a few things besides the main aim of this change

* utilizes typescript
* attempts to consolidate CLI colors giving them some specific meaning (this part is very very very opinionated, I need some feedback to understand if it makes sense to you too)
* updates the help with some more text


packages/contentful--create-contentful-app/src/index.ts
```diff
@@ -109,23 +88,34 @@ async function initProject(appName, options) {
...
+      code('  create-contentful-app init my-app --templateSource "github:user/repo"'),
+      '',
+      `Contentful official templates are hosted at ${highlight('https://github.com/contentful/apps/tree/master/templates')}.`
+    ].join('\n'))
+    .argument('[app-name]', 'Name of the app')
     .option('--npm', 'Use NPM')
     .option('--yarn', 'Use Yarn')
-    .option('--javascript, -js')
-    .option('--typescript, -ts')
+    .option('--javascript, -js', 'Use default javascript template')
```
colomolo, line 113:
```suggestion
    .option('--javascript, -js', 'Use default Javascript template')
```

packages/contentful--create-contentful-app/src/index.ts
```diff
@@ -109,23 +88,34 @@ async function initProject(appName, options) {
...
+    .option('--typescript, -ts', 'Use default typescript template')
```
colomolo, line 114:
```suggestion
    .option('--typescript, -ts', 'Use default Typescript template')
```

packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -0,0 +1,82 @@
...
+    let message = e.message ?? 'Unknown error';
+    if (e.code === 'DEST_NOT_EMPTY') {
+      message = 'Destination directory is not empty.';
+    }
+    throw new Error(message);
+  }
+}
+
+function validate(destination: string): void {
+  const packageJSONLocation = `${destination}/package.json`;
```
colomolo, line 44: Did I get it right — if we'll try to clone a non-existent folder, there will be no package.json?

shikaan: Exactly

colomolo: image

packages/contentful--create-contentful-app/src/logger.ts
```diff
@@ -0,0 +1,28 @@
...
+export function highlight(str: string) {
+  return chalk.bold(str);
+}
+
+export function choice(str: string) {
+  return chalk.greenBright(str);
+}
+
+export function success(str: string) {
+  return chalk.cyan(str);
```
colomolo, line 23: Maybe make success `greenBright` and 'choice' - cyan?

colomolo (approved):
Just one stylistic comment. Other than that LGTM 🔥

packages/contentful--create-contentful-app/src/logger.ts
shikaan: I have actually used the existing color scheme (which I am not sure how was chosen). Maybe @fabe has an idea or opinions?

colomolo: It was my who added greenBright as a fallback choice color, but that was pretty random.

shikaan: Maybe that can be addressed separately anyway. This branch includes a lot of changes already. WDYT?

packages/contentful--create-contentful-app/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+  "compilerOptions": {
```
andipaetzold, line 2: Can we just use `@tsconfig/node16` here instead. Thay way we are sure that we always have the correct `target` and `lib` set up: https://www.npmjs.com/package/@tsconfig/node16

packages/contentful--create-contentful-app/package.json
```diff
@@ -27,18 +26,23 @@
...
     "pre-commit": "lint-staged",
-    "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
+    "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
+    "build": "PARCEL_WORKERS=0 parcel build ./src/index.ts --dist-dir lib "
   },
   "dependencies": {
     "@contentful/app-scripts": "file:../contentful--app-scripts",
+    "@types/chalk": "^2.2.0",
+    "@types/node": "^17.0.17",
+    "@types/tildify": "^2.0.2",
```
andipaetzold, line 39: Those should be dev dependencies

packages/contentful--create-contentful-app/src/index.ts
```diff
@@ -109,23 +88,34 @@ async function initProject(appName, options) {
...
+      `Contentful official templates are hosted at ${highlight('https://github.com/contentful/apps/tree/master/templates')}.`
+    ].join('\n'))
+    .argument('[app-name]', 'Name of the app')
     .option('--npm', 'Use NPM')
     .option('--yarn', 'Use Yarn')
-    .option('--javascript, -js')
-    .option('--typescript, -ts')
+    .option('--javascript, -js', 'Use default Javascript template')
+    .option('--typescript, -ts', 'Use default Typescript template')
+    .option('-t, --templateSource <url>', [
```
andipaetzold, line 115: Not really sure I like these options. I would've assumed that we have a `--templates, -t` for directly referencing a template from the `apps` repo. And something short like `--src` for a full url.

packages/contentful--create-contentful-app/src/typings.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module '@contentful/app-scripts' {
+  const createAppDefinition: { interactive: () => Promise<void> };
+}
```
andipaetzold, line 3: Why is this necessary? These should come from `@contentful/app-scripts` itself

packages/contentful--create-contentful-app/package.json
andipaetzold: Also, `@types/chalk` and `@types/tildify` are a stub package

PR #873, opened 2022-02-15
ci: remove dependencies

shikaan:


packages/contentful--create-contentful-app/package.json
```diff
@@ -61,11 +58,13 @@
     "access": "public"
   },
   "devDependencies": {
+    "@tsconfig/node16": "^1.0.2",
+    "@types/chalk": "^2.2.0",
     "@types/degit": "^2.8.3",
     "@types/inquirer": "^8.2.0",
+    "@types/node": "^17.0.17",
     "@types/rimraf": "^3.0.2",
```
colomolo, line 66: Do we need `rimraf` at all?

shikaan: Yes, to perform the cleanup when the templating fails

colomolo (approved):

PR #875, opened 2022-02-16
feat: Add example option to app bootstrap

colomolo:


packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -3,24 +3,33 @@ import { existsSync, readFileSync } from 'fs';
...
+import { CLIOptions, ContentfulExample } from './types';
 import { highlight, warn } from './logger';
 import { rmIfExists } from './utils';
 
-function isContentfulTemplate(url: string) {
-  return Object.values(ContentfulTemplate).some(t => url.includes('contentful/apps/templates/' + t));
+const EXAMPLES_PATH = 'contentful/apps/examples/';
+
+function isContentfulExample(url: string) {
+  return Object.values(ContentfulExample).some(t => url.includes('contentful/apps/examples/' + t));
```
shikaan, line 13: maybe we can reuse the constant here too

packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -44,19 +53,19 @@ function validate(destination: string): void {
   const packageJSONLocation = `${destination}/package.json`;
   if (!existsSync(packageJSONLocation)) {
     throw new Error(
-      `Invalid template: missing "${packageJSONLocation}".`
```
shikaan, line 47: I would still call it template: the validation applies to examples (contentful things) and general templates

packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -65,15 +74,15 @@ function cleanUp(destination: string) {
   rmIfExists(resolve(destination, 'yarn.lock'));
 }
 
-export async function cloneTemplateIn(destination: string, options: CLIOptions) {
-  const source = getTemplateSource(options);
+export async function cloneExampleIn(destination: string, options: CLIOptions) {
```
shikaan, line 77: Also here I would use template, for the saem reason as above

colomolo: Yeah, makes sense.

shikaan (approved):
Looks good! 

There are a few references to examples instead of templates left in the `template.ts` file. The only place we should be using the word "example" is in `makeContentfulExampleSource` and the enum with the list of supported ones in my eyes

colomolo:
> There are a few references to examples instead of templates left in the `template.ts` file. The only place we should be using the word "example" is in `makeContentfulExampleSource` and the enum with the list of supported ones in my eyes

Thanks! Fixed.

PR #881, opened 2022-02-18
feat: React Apps Toolkit package [EXT-3489]

loweisz:
Added the new react-apps-toolkit package.
Mostly reused the things we have in the other packages as well.

I had to add the build command, as the hooks will be written in TS.

Also somehow the bootstrap command of lerna didn't install the packages for this project. Might be because of: https://github.com/lerna/lerna/issues/1457
that's why I changed it to the solution proposed in the issue.


shikaan (commented):
QQ: Why in master and not in canary?

andipaetzold:
> QQ: Why in master and not in canary?

@shikaan What would be the difference?

The first version of a new package is always tagged by `latest`. So keeping things in the `canary` branch does not really add much except that we might get more merge conflicts. I think it should be fine to develop directly on `master`. The package version is `0.0.x` and the README states that things are in development.

andipaetzold:
The first version of a package cannot be a canary/prerelease. Therefore merging this into master

andipaetzold:
Waiting with this PR until all packages are versioned independently

PR #882, opened 2022-02-18
chore: use node 16 & cimg images in CI

andipaetzold:
* We were still using node 12 for releases. Updated to node 16
* We used `node` images instead of the recommended `cimg/node` of CircleCI
* The `latest` is now `cimg/node:current`, so we don't need to update it in the future


.circleci/config.yml
```diff
@@ -12,12 +12,12 @@ install-deps: &install-deps
 executors:
   node-container-lts:
     docker:
-      - image: node:16
+      - image: cimg/node:16.13
```
kdamball, line 15: 16.14.0 is the lts, no? https://circleci.com/developer/images/image/cimg/node

andipaetzold: Correct, updated 👍

kdamball (approved):

PR #894, opened 2022-02-22
chore: use node lts 

kdamball:


.nvmrc
```diff
@@ -1 +1 @@
-v16.13.1
+v16.14.0
```
shikaan, line 1: Shall we make this v16 so it always picks latest LTS?

shikaan: (same above)

andipaetzold:
Why not just merge master?

.nvmrc
kdamball: Even better - it allows `lts` as an alias :)

shikaan (approved):

kdamball:
@andipaetzold we might also need this change for canary, right?

andipaetzold:
Yes, but why not merge master into canary? I did the same here https://github.com/contentful/create-contentful-app/pull/882

PR #909, opened 2022-03-01
feat: [EXT-3490] add useSDK hook

loweisz:
Adding the hook and the provider

### DIFF
--- a/packages/contentful--react-apps-toolkit/package.json
+++ b/packages/contentful--react-apps-toolkit/package.json
@@ -46,6 +46,6 @@
     "access": "public"
   },
   "dependencies": {
-    "@contentful/app-sdk": "^4.3.5"
+    "@contentful/app-sdk": "^4.0.0"
   }
 }
--- a/packages/contentful--react-apps-toolkit/package.json
+++ b/packages/contentful--react-apps-toolkit/package.json
@@ -31,7 +31,7 @@
     "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
   },
   "peerDependencies": {
-    "@contentful/app-sdk": ">=4.3.5",
+    "@contentful/app-sdk": ">=4.0.0",
     "react": ">=16.8.0"
   },
   "devDependencies": {
@@ -45,8 +45,5 @@
   },
   "publishConfig": {
     "access": "public"
-  },
-  "optionalDependencies": {
-    "@contentful/f36-components": "^4.0.35"
   }
 }
--- a/packages/contentful--react-apps-toolkit/package.json
+++ b/packages/contentful--react-apps-toolkit/package.json
@@ -31,7 +31,6 @@
     "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
   },
   "peerDependencies": {
-    "@contentful/app-sdk": ">=4.0.0",
     "react": ">=16.8.0"
   },
   "devDependencies": {
@@ -45,5 +44,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "dependencies": {
+    "@contentful/app-sdk": "^4.3.5"
   }
 }
--- a/packages/contentful--react-apps-toolkit/package.json
+++ b/packages/contentful--react-apps-toolkit/package.json
@@ -31,9 +31,8 @@
     "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
   },
   "peerDependencies": {
-    "react": ">=16.8.0",
     "@contentful/app-sdk": ">=4.3.5",
-    "@contentful/f36-components": ">=4.0.0"
+    "react": ">=16.8.0"
   },
   "devDependencies": {
     "@testing-library/react": "^12.1.3",
@@ -46,5 +45,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "optionalDependencies": {
+    "@contentful/f36-components": "^4.0.35"
   }
 }
--- a/packages/contentful--react-apps-toolkit/package.json
+++ b/packages/contentful--react-apps-toolkit/package.json
@@ -31,7 +31,9 @@
     "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
   },
   "peerDependencies": {
-    "react": ">=16.8.0"
+    "react": ">=16.8.0",
+    "@contentful/app-sdk": ">=4.3.5",
+    "@contentful/f36-components": ">=4.0.0"
   },
   "devDependencies": {
     "@testing-library/react": "^12.1.3",
--- a/packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
+++ b/packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
@@ -1,4 +1,4 @@
-import React, { FC, ReactElement, useRef } from 'react';
+import React, { FC, ReactElement } from 'react';
 import { init, KnownSDK } from '@contentful/app-sdk';
 
 export const SDKContext = React.createContext<{ sdk: KnownSDK | null }>({ sdk: null });
--- a/packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
+++ b/packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
@@ -1,6 +1,5 @@
 import React, { FC, ReactElement } from 'react';
 import { init, KnownSDK } from '@contentful/app-sdk';
-import { Spinner } from '@contentful/f36-components';
 
 export const SDKContext = React.createContext<{ sdk: KnownSDK | null }>({ sdk: null });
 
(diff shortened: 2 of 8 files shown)


kdamball (approved):

packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const SDKProvider: FC<SDKProviderProps> = (props) => {
+  const [sdk, setSDK] = React.useState<KnownSDK | undefined>();
+
+  React.useEffect(() => {
+    init(setSDK);
+  }, []);
+
+  if (!sdk) {
+    return props.loading ?? <Spinner />;
```
colomolo, line 19: Do we really want to block everything while sdk is loading? Maybe at least create escape hatch - when no `loading` prop is provided, we pass `undefined` sdk and let handle loading inside of the app? Wdyt?

packages/contentful--react-apps-toolkit/package.json
```diff
@@ -31,7 +31,9 @@
     "commit-msg": "commitlint -E HUSKY_GIT_PARAMS"
   },
   "peerDependencies": {
-    "react": ">=16.8.0"
+    "react": ">=16.8.0",
+    "@contentful/app-sdk": ">=4.3.5",
+    "@contentful/f36-components": ">=4.0.0"
```
andipaetzold, line 36: I think this is better, right?

Also, why make f36 a peer dep?

```suggestion
    "@contentful/app-sdk": "^4.3.5",
    "@contentful/f36-components": "^4.0.0"
```

andipaetzold: Can we just depend on `@contentful/f36-spinner`? We don't need the full repo. This would also make the bundle smaller or the treeshaking easier.

packages/contentful--react-apps-toolkit/src/useSDK.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+import { useSDK } from './index';
+import ShallowRenderer from 'react-test-renderer/shallow';
+import React from 'react';
+
+let realUseContext: any;
+let useContextMock: any;
+// Setup mock
+beforeEach(() => {
+  realUseContext = React.useContext;
+  useContextMock = React.useContext = jest.fn().mockReturnValue({ sdk: 'mocked-sdk' });
```
andipaetzold, line 11: Why not properly mock `useContext`?

packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
andipaetzold, line 19: Otherwise we return `false`

```suggestion
    return props.loading ? <Spinner /> : null;
```

andipaetzold (changes requested):
Can we add some jsdoc for the hook and the component?

packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
loweisz: I personally think I would prefer to have always an SDK returned from the app, otherwise, the user always needs to add a check for the existence of it. I think that was also the initial plan or not? 🤔

cc @andipaetzold

packages/contentful--react-apps-toolkit/package.json
loweisz: I would expect that f36 is more likely to be already installed in the app anyway already, that's why I choose the whole library. 
I honestly think we could also get rid of the spinner and just add a loading string or just return null, wdyt?

loweisz:
> Can we add some jsdoc for the hook and the component?

What would be the benefit of it additionally to the ts types ? 🤔

packages/contentful--react-apps-toolkit/package.json
andipaetzold: Not having forma 36 as a dependency would be great, yes. We could make it an optional dependency and if it's installed, we render the spinner?

`@contentful/f36-spinner` is a dependency `@contentful/f36-components`, so I'd always just directly use that to reduce the node_modules folder. Some users might not want to install all of forma 36.

packages/contentful--react-apps-toolkit/src/useSDK.spec.ts
loweisz: I didn't want to add jsx stuff here, so i prefered this way as I only wanted to test the hook functionality and not the context

andipaetzold: I thought about

```
jest.mock('react', () => {
  return {
    ...jest.requireActual('react'),
    useContext: () => ({ sdk: 'mocked-sdk' })
  }
});
```

That way you would not manipulate a dependency for your test. I see the risk that a failing test might also break other tests.

andipaetzold:
> > Can we add some jsdoc for the hook and the component?
> 
> What would be the benefit of it additionally to the ts types ? 🤔

The docs of `useSDK` would explain that you can only use it when it's wrapped in an `SDKProvider`. And explain that depending on the locaiton you can specify the type of the returned sdk.

packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
colomolo: ok 👍

packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+interface SDKProviderProps {
+  loading?: ReactElement;
+}
+
+export const SDKProvider: FC<SDKProviderProps> = (props) => {
+  const [sdk, setSDK] = React.useState<KnownSDK | undefined>();
+
+  React.useEffect(() => {
+    init(setSDK);
```
Jwhiles, line 15: I know it's unlikely, but there's a chance that something goes wrong in init and the callback is never called. Do we want to have some sort of timeout here to let the user know something's gone wrong after a certain time has elapsed?

packages/contentful--react-apps-toolkit/src/useSDK.spec.ts
loweisz: Ah yes, that's better I agree 👍

packages/contentful--react-apps-toolkit/package.json
loweisz: Good idea, I've added it now as an optional dependent and render nothing if it is not available

andipaetzold (changes requested):
~* When adding a peer or optional dependency, it must be. installed as dev dependency as well. Otherwise, it won't be available during development~
* Increase the version range to their maximum: Forma36 v4.0.0. App SDK all since CMA adapter was introduced
~* We should still replace `f36-components` with `f36-spinner`~
~* Just noticed that there is also the concept of optional peer deps, which looks more what we want (sorry for the back and forth)~
~* We need to handle optional dependencies properly: https://docs.npmjs.com/cli/v8/configuring-npm/package-json#optionaldependencies~

andipaetzold:
We decided to drop the spinner as it probably wouldn't even be rendered.

packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FC, ReactElement } from 'react';
+import React, { FC, ReactElement, useRef } from 'react';
```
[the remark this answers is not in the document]
loweisz: https://github.com/contentful/create-contentful-app/pull/909/commits/52b036fe120dda21c5bf1a779c5e479dcff8448d 🤫

packages/contentful--react-apps-toolkit/src/useSDK.ts
```diff
@@ -0,0 +1,20 @@
+import { KnownSDK } from '@contentful/app-sdk';
+import { useContext } from 'react';
+import { SDKContext } from './SDKProvider';
+
+/**
+ * A react hook providing the AppSdk.
+ *  - The type of sdk varies depending on the location where it is used.
+ *  - As it is depending on the context providing the sdk, the hook can only be used within the SDKProvider
+ */
```
andipaetzold, line 9:
```suggestion
/**
 * A react hook returning the App SDK.
 *  - The type of SDK varies depending on the location where it is used.
 *  - As it is depending on the context providing the SDK, the hook can only be used within the SDKProvider
 */
```

packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface SDKProviderProps {
+  loading?: ReactElement;
+}
+
+/**
+ * The Component providing the AppSdk, the useSDK hook can only be used within this Provider
+ * @param props
+ * @constructor
+ */
```
andipaetzold, line 14:
```suggestion

/**
 * The Component providing the AppSdk, the useSDK hook can only be used within this Provider
 * 
 * @param props
 * @constructor
 */
```

andipaetzold: Why `@constructor`? Can you add a description for props?

andipaetzold (approved):

Jwhiles (approved):

PR #910, opened 2022-03-01
fix: [EXT-3565] update node support

martin3walker:


packages/contentful--create-contentful-app/package.json
```diff
@@ -10,7 +10,7 @@
     "directory": "packages/contentful--create-contentful-app"
   },
   "engines": {
-    "node": ">=12",
+    "node": ">=14.14.0",
```
andipaetzold, line 13: Why starting at 14.14 and not 14.15?

.gitignore
```diff
@@ -23,3 +23,4 @@ dist
 .idea
 .vscode
 .parcel-cache
+.history
```
andipaetzold, line 26: What's that about?

packages/contentful--create-contentful-app/.npmrc
```diff
@@ -0,0 +1 @@
+engine-strict=true
```
andipaetzold, line 1: Good way to avoid using unsupported dependencies 👍

packages/contentful--create-contentful-app/package-lock.json
```diff
@@ -1,1835 +1,8 @@
 {
   "name": "@contentful/create-contentful-app",
   "version": "1.1.0-alpha.1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 1,
```
andipaetzold, line 4: Please revert me :)

andipaetzold (changes requested):
Can you update the README and `create-contentful-app` as well?

packages/contentful--create-contentful-app/package.json
martin3walker: I thought I'd successfully run the command with node 14.14, but I just double checked and that isn't true.

packages/contentful--create-contentful-app/package-lock.json
martin3walker: Any idea how I managed to do this?

packages/contentful--create-contentful-app/package.json
martin3walker: Right, which you were saying is a good proxy for what versions of node are being used, and thus what we should support. I remember now, thank you!

andipaetzold:
Let's adjust the CircleCI config to match the node versions and we are good to go 🚀

andipaetzold (approved):

PR #911, opened 2022-03-02
docs: [] update --help

shikaan:


andipaetzold (approved):
Should be a `fix:` as this is a change in the code and actually something broken

PR #916, opened 2022-03-03
chore: cli tweaks

colomolo:


packages/create-contentful-app/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "create-contentful-app",
   "version": "1.1.0-alpha.3",
   "description": "A template for building Contentful Apps",
+  "keywords": ["Contentful", "create-contentful-app", "CCA"],
```
shikaan, line 5: @stefanjudis Do you happen to have a list of good keywords we should be using here?

packages/contentful--create-contentful-app/README.md
```diff
@@ -5,7 +5,7 @@ This project makes it easy to bootstrap [Contentful Apps](https://www.contentful
 ## Requirements
 
 - Node.js v14 or v16
-- NPM v6
+- npm v6
```
shikaan, line 8: I think this was meant to be changed in the --help too. We also need to review all the links there, I fixed one yesterday, not sure about the rest

colomolo: yep, I changed it the help command too. Which links do you mean?

packages/contentful--create-contentful-app/src/logger.ts
```diff
@@ -16,13 +16,13 @@ export function highlight(str: string) {
...
 }
 
 export function success(str: string) {
-  return chalk.cyan(str);
+  return chalk.greenBright(str);
 }
 
 export function code(str: string) {
-  return chalk.dim(str);
+  return chalk.bold(str);
```
colomolo, line 27: Color changes were approved by @fabe

packages/contentful--create-contentful-app/README.md
shikaan: https://github.com/contentful/create-contentful-app/blob/f338b5a609b8bdba1d9095a399abc8974025a723/packages/contentful--create-contentful-app/src/index.ts#L132

This is still uppercase ☝️  Also, `Yarn -> yarn` maybe

Links are already updated, you are right :)

packages/create-contentful-app/package.json
stefanjudis: I'd drop `CCA`. Primarily `contentful` (I'd go with a lower `c`) is important I'd say. :)

stefanjudis (approved):
🎊 🚢

shikaan (approved):

PR #917, opened 2022-03-03
feat: [EXT-3491] add useCMAClient hook

martin3walker:


packages/contentful--react-apps-toolkit/package.json
```diff
@@ -46,6 +46,8 @@
     "access": "public"
   },
   "dependencies": {
-    "@contentful/app-sdk": "^4.0.0"
+    "@contentful/app-sdk": "^4.0.0",
+    "contentful-management": "^8.2.0",
+    "react": "^17.0.2"
```
andipaetzold, line 51: `react` is a peer dependency. Please install it as a dev dependency as well.

packages/contentful--react-apps-toolkit/src/useCMAClient.ts
```diff
@@ -0,0 +1,36 @@
+import { createClient, PlainClientAPI } from "contentful-management";
+import { useEffect, useState } from "react";
+import { useSDK } from "./index";
+
+/**
+ * A react hook returning the plain client CMA.
+ *  - Depends on Apps SDK.
+ *  - Must be called in the SDKProvider component (./SDKProvider). If called outside the provider, will throw error. useSDK hook called on line 12 handles error.
```
shikaan, line 8: What about being deliberate, throw and exception and tell the user what's up? 
I think it's not ideal development experience if you get the error and you need to look through the source why. WDYT?

packages/contentful--react-apps-toolkit/package.json
martin3walker: I believe this is fixed now.

packages/contentful--react-apps-toolkit/src/useCMAClient.ts
martin3walker: The `useSDK` hook would throw an exception first if this hook is not wrapped in the `SDKProvider`. So it didn't seem to make sense to add an error here. But, now that you mention it, the error being thrown would be specific to the `useSDK` hook, which is confusing. So maybe we can make it more general. Like "Make sure this hook is being used inside of the SDKProvider".

andipaetzold (changes requested):

packages/contentful--react-apps-toolkit/package.json
```diff
@@ -46,6 +46,8 @@
     "access": "public"
   },
   "dependencies": {
-    "@contentful/app-sdk": "^4.0.0"
+    "@contentful/app-sdk": "^4.0.0",
+    "contentful-management": "^8.2.0",
```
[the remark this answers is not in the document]
martin3walker: I think I'm confused. I thought `^4.0.0` meant everything above 4.0.0 is supported?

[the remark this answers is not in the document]
andipaetzold: This comment is about `contentful-management`. You currently require the latest version. CMA adapter was introduced in 7.13.0

packages/contentful--react-apps-toolkit/src/useCMA.ts
```diff
@@ -0,0 +1,32 @@
...
+        type: "plain",
+        defaults: {
+          environmentId: sdk.ids.environment,
+          spaceId: sdk.ids.space,
+        },
+      }
+    );
+
+    setCMA(client);
+  }, [sdk]);
```
andipaetzold, line 29: Can we use `useMemo` here? No need for `useEffect`

packages/contentful--react-apps-toolkit/package.json
```diff
@@ -48,6 +48,6 @@
   },
   "dependencies": {
     "@contentful/app-sdk": "^4.0.0",
-    "contentful-management": "^8.2.0"
+    "contentful-management": "^7.13.0"
```
andipaetzold: This doesn't include v8

So we should do `^7.13.0 || ^8.0.0` or `>= 7.13.0`

andipaetzold (changes requested):
Last two comments. We should be good after that

packages/contentful--react-apps-toolkit/src/useCMAClient.ts
martin3walker: @shikaan If either hook is used outside of the provider, an error saying "Make sure this hook is used inside the SDKProvider" will be thrown. WDYT?

shikaan (approved):

packages/contentful--react-apps-toolkit/src/useCMA.ts
martin3walker: Implemented with use memo, but maybe have a quick look to make sure I've done it right :)

andipaetzold (approved):
Made a tiny doc change. But looks good :)

PR #918, opened 2022-03-03
feat: [] add dev showcase to app scripts

loweisz:
Adding a success message to the upload script linking to the dev showcase.

Screen Shot 2022-03-03 at 15 21 50

**NOTE:** This can only be merged when the link is ready


loweisz:
> * https://ctfl.io/submit-app

Great thank you. Would be small enough: 
Screen Shot 2022-03-03 at 15 41 28

fabe:
Maybe just put the links in parenthesis and go without the line break?

loweisz:
> Maybe just put the links in parenthesis and go without the line break?

Like this:
Screen Shot 2022-03-03 at 15 53 26

fabe (approved):
Looks good, I'd just remove the `:` then!

shikaan:
How do we feel about adding the message also when you publish a new bundle? 

I think it's unlikely you want to put an app in the showcase when it's just been created (might be empty or still not ready to be shared), whereas the promotion of a bundle seems more likely to be a good time for that. WDYT? @fabe @loweisz

PR #919, opened 2022-03-03
feat: [] add timeout and console warn

loweisz:
Adding a timeout to the init function in the SDKProvider

### DIFF
--- a/packages/contentful--react-apps-toolkit/jest.config.js
+++ b/packages/contentful--react-apps-toolkit/jest.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   testEnvironment: 'jest-environment-jsdom',
   transform: {
-    '^.+\\.ts$': 'ts-jest',
+    '^.+\\.t(s|sx)$': 'ts-jest',
   },
-  testRegex: '^.+\\.spec\\.ts$'
+  testRegex: '^.+\\.spec\\.t(s|sx)$',
 };
--- a/packages/contentful--react-apps-toolkit/jest.config.js
+++ b/packages/contentful--react-apps-toolkit/jest.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   testEnvironment: 'jest-environment-jsdom',
   transform: {
-    '^.+\\.t(s|sx)$': 'ts-jest',
+    '^.+\\.tsx?$': 'ts-jest',
   },
-  testRegex: '^.+\\.spec\\.t(s|sx)$',
+  testRegex: '^.+\\.spec\\.tsx?$',
 };
--- a/packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
+++ b/packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
@@ -1,4 +1,4 @@
-import React, { FC, ReactElement } from 'react';
+import React, { FC, ReactElement, useRef } from 'react';
 import { init, KnownSDK } from '@contentful/app-sdk';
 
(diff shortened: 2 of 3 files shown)


packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
```diff
@@ -7,15 +7,35 @@ interface SDKProviderProps {
...
 
   React.useEffect(() => {
-    init(setSDK);
+    timeout.current = window.setTimeout(() => {
+      console.warn(
+        'Your app is taking longer than expected to initialize. If this is not fixed, let us know: https://github.com/contentful/ui-extensions-sdk/issues'
+      );
+    }, DELAY_TIMEOUT);
+    init(onInitSuccess);
+    return () => clearPossibleTimeout();
```
andipaetzold, line 38: Your solution feels quite verbose to me. Can't we just do this and remove the outside methods and imports? There is no need to store the timeout on component scope.

```suggestion
    const timeout = window.setTimeout(() => {
      console.warn(
        'Your app is taking longer than expected to initialize. If this is not fixed, let us know: https://github.com/contentful/ui-extensions-sdk/issues'
      );
    }, DELAY_TIMEOUT);
    init((sdk) => {
      setSDK(sdk);
      clearTimeout(timeout);
    });
    return () => clearTimeout(timeout);
```

andipaetzold (commented):
Can we add a tests to avoid accidentally showing wrong warnings?

packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
```diff
@@ -7,15 +7,35 @@ interface SDKProviderProps {
...
+  const onInitSuccess = (sdk: KnownSDK) => {
+    setSDK(sdk);
+    clearPossibleTimeout();
+  };
 
   React.useEffect(() => {
-    init(setSDK);
+    timeout.current = window.setTimeout(() => {
+      console.warn(
+        'Your app is taking longer than expected to initialize. If this is not fixed, let us know: https://github.com/contentful/ui-extensions-sdk/issues'
```
shikaan, line 34: The error in this case might still be in the user's code. What about rephrasing this making clear we won't do code reviews? :P 

```suggestion
        'Your app is taking longer than expected to initialize. If you think this is an error with Contentful's App SDK, let us know: https://github.com/contentful/ui-extensions-sdk/issues'
```

packages/contentful--react-apps-toolkit/src/SDKProvider.tsx
shikaan: Wouldn't you get the `console.warn` in case of a re-render where one of the new renders is successful?

loweisz: You are right andi, since we don't use the timeout outside of the useEffect we can just keep it there👍

shikaan (approved):

andipaetzold (approved):

packages/contentful--react-apps-toolkit/jest.config.js
```diff
@@ -1,7 +1,7 @@
 module.exports = {
   testEnvironment: 'jest-environment-jsdom',
   transform: {
-    '^.+\\.ts$': 'ts-jest',
+    '^.+\\.t(s|sx)$': 'ts-jest',
```
andipaetzold, line 4: Nit: the elegant way is `tsx?`

andipaetzold (approved):

PR #923, opened 2022-03-08
feat: useFieldValue [EXT-3492]

andipaetzold:
No update to the README as this will be done in another PR

### DIFF
--- a/packages/contentful--react-apps-toolkit/src/useFieldValue.tsx
+++ b/packages/contentful--react-apps-toolkit/src/useFieldValue.tsx
@@ -1,3 +1,4 @@
+import { EntryFieldAPI, KnownSDK } from '@contentful/app-sdk';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useSDK } from './useSDK';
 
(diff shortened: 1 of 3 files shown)


packages/contentful--react-apps-toolkit/src/useFieldValue.tsx
```diff
@@ -0,0 +1,81 @@
...
+      throw new Error(
+        '`useFieldValue` can only be used in field, sidebar or entry editor location'
+      );
+    }
+
+    const fieldIdWithDefault = fieldId ?? ('field' in sdk ? sdk.field.id : undefined);
+
+    if (!fieldIdWithDefault) {
+      throw new Error(
+        'Missing `fieldId`. The `fieldId` can only be omitted when your app is renderd in field location.'
```
Jwhiles, line 49: I am mildly opposed to this part of design. I would prefer a `useFieldValue` and `useEntryFieldValue`/`useFieldValueById` hook which would handle the two different use cases that this hook currently does.

I'd prefer this because I think it's slightly simpler. I find overloading functions like this a bit confusing.

andipaetzold: Is the issue about the hook having a different behavior in different locations? Or that you have to pass in an object instead of string arguments?

Jwhiles: My object is that this one hook tries to handle two different use cases. I don't think it's a big issue, and a lot of our existing APIs are overloaded in this way, but I do think it's a bit confusing, and probably harder to maintain than just having two hooks.

andipaetzold: As a user of the package, I think it is good to only have one hook that you can import. Even when we split things up and export two hooks, people will probably import the wrong one.

I think I can improve the current implementation, so both use cases are not mixed that much. (extract an internal hook with a mandatory fieldId)

Jwhiles (approved):

PR #962, opened 2022-03-28
[EXT-3604] Update React Hooks readme to match public docs

martin3walker:


shikaan (commented):
LGTM, why have we moved links from npm to github?

martin3walker:
It might make more sense just to link the public docs in this repo. Otherwise, we have to update the readme every time we update the docs. Thoughts @andipaetzold?

andipaetzold:
> It might make more sense just to link the public docs in this repo. Otherwise, we have to update the readme every time we update the docs. Thoughts @andipaetzold?

I don't understand why we would need to update the links when the docs change?

In general, I would prefer linking to npm as that's more likely where people come from and will read the CCA docs anyway. But that's just a guess.

martin3walker:
> I don't understand why we would need to update the links when the docs change?
I mean rather than copy pasting the doc app content into the readme, let's just link to that article. That way we only have to keep one thing up to date.

andipaetzold:
I, as a developer, prefer if all necessary information is available in the README formatted using simple markdown. I think that has value over linking to an external page.

andipaetzold (approved):

PR #979, opened 2022-04-05
feat: return updated value when setting a field value via `useFieldValue` [EXT-3654]

andipaetzold:
Sometimes the value changes when calling `setValue`. With this change, the value is returned when calling `setValue` of `useFieldValue`.

### DIFF
--- a/packages/contentful--react-apps-toolkit/src/useFieldValue.tsx
+++ b/packages/contentful--react-apps-toolkit/src/useFieldValue.tsx
@@ -1,10 +1,10 @@
-import { EntryFieldAPI, KnownSDK } from '@contentful/app-sdk';
+import { EntryFieldAPI, KnownSDK, SerializedJSONValue } from '@contentful/app-sdk';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import { useSDK } from './useSDK';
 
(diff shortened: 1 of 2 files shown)


packages/contentful--react-apps-toolkit/src/useFieldValue.tsx
```diff
@@ -34,7 +34,7 @@ export function useFieldValue<Value = unknown>(
   const updateValue = useCallback(
     async (newValue: Value | undefined) => {
       setValue(newValue);
-      await entryFieldApi.setValue(newValue, localeWithDefault);
```
loweisz, line 37: can we not just return without the await and remove the `async` from the function?

Jwhiles (approved):

loweisz (approved):

packages/contentful--react-apps-toolkit/src/useFieldValue.tsx
andipaetzold: See this example. Removing the `await` from `A()` will remove `A` from the stack tracek

```js
async function A() {
  return await B();
}

async function B() {
  await Promise.resolve();
  throw new Error('BEEP BEEP');
}

A();

```

PR #987, opened 2022-04-07
chore: explicitly add react-test-renderer as dev dependency

andipaetzold:
Required by testing library. I don't know why it currently works, but it is failing in `canary`.
```
Could not auto-detect a React renderer. Are you sure you've installed one of the following
      - react-dom
      - react-test-renderer
    If you are using a bundler, please update your imports to use a specific renderer.
    For instructions see: https://react-hooks-testing-library.com/installation#being-specific
```


loweisz:
Why not the test-renderer? We are not really rendering something here, so maybe this could be a bit more lightweight

andipaetzold:
No real reason, switching then

loweisz (approved):

PR #994, opened 2022-04-08
chore: ignore package-lock.json changes

andipaetzold:
This should reduce the number of releases.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -5,8 +5,14 @@
   "command": {
     "publish": {
       "ignoreChanges": [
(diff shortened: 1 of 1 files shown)


shikaan (approved):
Should we add also other non-feature changes here? Like test files for example

andipaetzold:
Sounds good. I'll go through the list of files and update the PR.

andipaetzold:
I added some more

shikaan (approved):

PR #1053, opened 2022-05-10
fix: [] remove app-sdk check

loweisz:
To enable also non frontend template we remove the hard check for the app-sdk. 

I thought about adding a pre check to find out if it is a frontend app. But I could not find a reasonable way to find that out, so I decided to just remove it for now

### DIFF
--- a/packages/contentful--create-contentful-app/src/template.ts
+++ b/packages/contentful--create-contentful-app/src/template.ts
@@ -10,7 +10,7 @@ import { rmIfExists } from './utils';
 const EXAMPLES_PATH = 'contentful/apps/examples/';
 
 function isContentfulTemplate(url: string) {
-  return Object.values(ContentfulExample).some(t => url.includes(EXAMPLES_PATH + t));
+  return Object.values(ContentfulExample).some((t) => url.includes(EXAMPLES_PATH + t));
 }
 
 function makeContentfulExampleSource(options: CLIOptions) {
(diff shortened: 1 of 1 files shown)


packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -48,28 +48,21 @@ async function clone(source: string, destination: string) {
       // to avoid confusing people.
       throw new Error('Destination directory is not empty.');
     }
-    throw e
+    throw e;
   }
 }
 
 function validate(destination: string): void {
```
andipaetzold, line 55: Can't we remove the whole function?

loweisz: I think it might be helpful to validate if it holds a valid `package.json`. wdyt?

andipaetzold: We will remove that part then as soon as we have an example with both frontend & backend in different folders

loweisz: I guess it depends on the template we build the, but all of the apps we built with frontend&backend have a package.json at root level, so probably this template will also have one

I just think we could prevent some gotchas when people are building custom templates

andipaetzold (approved):

andipaetzold:
That's more a `fix` or `feat`, but not `refactor`. Otherwise we won't release a new version

PR #1095, opened 2022-06-14
docs: fix example for useCMA

backflip:
After following https://www.contentful.com/developers/docs/extensibility/app-framework/create-contentful-app/ to create a new app, I stumbled upon `cma.entries` being `undefined`. Additionally, `getMany()` returns a `CollectionProp`, so for the demo we are probably interested in its `items` property.


loweisz (approved):
Great catch 💯  Looks good to me, thanks for the contribution

PR #1255, opened 2022-11-05
feat: add host argument for bundle cleanup

ivo-toby:


packages/contentful--app-scripts/lib/get-management-token.js
```diff
@@ -5,30 +5,27 @@ const open = require('open');
...
-
-  } catch(err) {
+  } catch (err) {
     return false;
   }
-}
+};
 
-async function getManagementToken() {
+async function getManagementToken(host = undefined) {
```
vikaskumr, line 20: one small question, why to specifically define `undefined` as by default parameters have `undefined` value in javascript.

ivo-toby: sematics; it's an optional argument.. in Typescript i'd use `?` which is way more clear, but I've removed it

vikaskumr (approved):
Nice work!

PR #1267, opened 2022-11-16
chore: bump lerna@6 and enable Nx on CircleCI [NONE]

marcolink:
## Description
Since `lerna@5.1` it s possible to use Nx for task scheduling which results in faster build times by only execute changed packages. Starting with `lerna@6` build optimization is enabled by default through [Nx](https://nx.dev/).
### What changes
- Ensure latest version of lerna (`6.0.3`)
- Enable Nx on `lerna.json`
- Ensure existence of `nx.json` (can be used for further optimization by defining dependant tasks). Please feel free to further adjust to your needs.
- Add [Nx Orb](https://github.com/nrwl/nx-orb) to only execute "affected" commands. (This requires further configuration - please contact @vikaskumr for questions)
### Expected outcome
Decrease local build time and execution time on CircleCI
### Further links
- [Integrating Nx and Lerna](https://nx.dev/recipes/adopting-nx/lerna-and-nx)
- [Speeding up computations in GitHub/CircleCI workflows with Affected Commands](https://blog.nrwl.io/using-nx-affected-on-ci-d3d2686295a0)


package.json
```diff
@@ -32,6 +32,7 @@
     "lerna": "6.0.3",
     "lint-staged": "13.0.3",
     "mocha": "10.1.0",
+    "nx": "^15.1.0",
```
andipaetzold, line 35: Please pin

.circleci/config.yml
```diff
@@ -1,128 +1,127 @@
...
+          template-preset: npm-read
       - run: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
-      - run: git config --global user.email "tech-accounts+gh+whydah-gally@contentful.com"
+      - run: git config --global user.email
+          "tech-accounts+gh+whydah-gally@contentful.com"
       - run: git config --global user.name "whydah-gally"
-      - *install-deps
-      - *build
-
+      - *a2
```
andipaetzold, line 83: Why are you changing the names `install-deps` and `build` to `a2` and `a3`?

andipaetzold (changes requested):
Any reason for reformatting the whole circle ci config? Do we now have a formatting guideline?

.circleci/config.yml
marcolink: Unfortunately the yml formatter has it's own opinions - fixed!

nx.json
```diff
@@ -0,0 +1,29 @@
...
+    "default": {
+      "runner": "nx/tasks-runners/default",
+      "options": {
+        "cacheableOperations": [
+          "dist",
+          "build",
+          "test",
+          "lint",
+          "test:ci",
+          "tsc"
```
andipaetzold, line 12: Where do these operations come from? They don't match what we have within this repo.

nx.json
```diff
@@ -0,0 +1,29 @@
...
+      }
+    }
+  },
+  "targetDefaults": {
+    "build": {
+      "dependsOn": [
+        "^build"
+      ]
+    },
+    "dev": {
```
andipaetzold, line 23: We don't have a `dev` target/script

nx.json
marcolink: This is the set of "default" commands which should be adjusted to your needs and situation. We did not want to just populate the list with all your available script commands to avoid unexpected behavior.

andipaetzold: Okay, I'll update the file then.
Thanks

andipaetzold (approved):

PR #1277, opened 2022-11-22
fix: get the correct amount of bundles to delete when using cleanup-b…

ivo-toby:
The reported issue occurred when deleting large amounts of bundles (> 1000). This fix will allow large amounts to be deleted as well.


packages/contentful--app-scripts/lib/clean-up/clean-up-bundles.js
```diff
@@ -29,20 +32,62 @@ const scheduleBundleDeletion = async (bundlesToDelete, client, settings) => {
 };
 
 async function cleanUpBundles(settings) {
```
colomolo, line 34: I wonder if it's possible to break up this function a little bit as it's really big now and includes few other functions in it.

packages/contentful--app-scripts/lib/clean-up/clean-up-bundles.js
```diff
@@ -29,20 +32,62 @@ const scheduleBundleDeletion = async (bundlesToDelete, client, settings) => {
...
+  const bundlesSpinner = ora(`Fetching all bundles...`).start();
+  try {
+    const { total } = await getAppBundleCount();
+    const amountToDelete = total - settings.keep;
+    if (amountToDelete < 1) {
+      console.log(`${chalk.yellow('Warning:')} There is nothing to delete`);
+      bundlesSpinner.stop();
+      return;
+    }
+    bundlesToDelete = await getAppBundles(amountToDelete);
```
marcolink, line 89: As a quick fix, we could just sort them by their `sys.createdAt` information - wdyt?

ivo-toby: I tried that but it seems the sorting didn't work. I have a call to check this scheduled today

marcolink: Ah, what I meant is, sorting your local list, since you're fetching all bundles, you can sort them locally, and just delete the oldest bundles.

packages/contentful--app-scripts/lib/clean-up/clean-up-bundles.js
ivo-toby: I can of course, though I'd rather clean up the whole thing and rewrite it to TS with proper modules.

colomolo (approved):

packages/contentful--app-scripts/lib/clean-up/clean-up-bundles.js
```diff
@@ -29,20 +32,67 @@ const scheduleBundleDeletion = async (bundlesToDelete, client, settings) => {
...
+      if (currLength <= requestedAmount && result.total > currLength) {
+        return [...result.items, ...(await getBundles(result.items.length, currLength))];
+      } else {
+        return result.items;
+      }
+    };
+
+    const all = await getBundles(DEFAULT_BUNDLES_TO_FETCH, 0);
+
+    return all.slice(0, requestedAmount);
```
marcolink, line 73: Shouldn't we only slice after sorting the list?

ivo-toby: hmm... Yeah. That's why I wanted the sorting from the API in the first place :D 
fix coming up

marcolink (approved):
LGTM, do we have a test that ensures, we're always deleting the oldest bundles?

ivo-toby:
> 

I'll add a test for that after my call about the API sorting.

ivo-toby:
@marcolink the API does not support sorting from the `order`-param, it's hardcoded sorting on `createdAt`. 
Because I want to be sure I have the oldest possible bundles I fetch them all instead of a limited list and then reverse and slice to have them removed. There's no integration testing in place atm, I'll discuss that with the team.

colomolo (approved):
Looks good 👍

PR #1284, opened 2022-11-28
fix: custom host parameter for cleanup [HOMER-1343]

Chaoste:
- fix: support host parameter in CI
- test: add test cases for custom host and complete bundle fetching


ivo-toby (approved):
likeitalot

PR #1294, opened 2022-12-01
fix: Update CMA client type []

colomolo:
Updating types for plain client. Changes:
https://github.com/contentful/ui-extensions-sdk/pull/1430

`CMAClient` is only available from App SDK v4.14.0, but we want to keep compatible with old versions
If `CMAClient` is available, `useCMA` should return `CMAClient`
If `CMAClient` is not available, `useCMA` should return `PlainClientAPI`
Source: https://stackoverflow.com/a/61625831/17270873


packages/contentful--react-apps-toolkit/src/useCMA.ts
```diff
@@ -1,12 +1,13 @@
-import { createClient, PlainClientAPI } from 'contentful-management';
+import { createClient } from 'contentful-management';
+import { CMAClient } from '@contentful/app-sdk';
 import { useMemo } from 'react';
 import { useSDK } from './useSDK';
 
 /**
  * React hook returning a CMA plain client instance.
```
andipaetzold, line 7: Should we mention that the type differs from `PlainClientAPI`? Not quite sure - also fine with leaving it as it is.

andipaetzold (approved):

colomolo:
LGTM 👍 (can't approve)

PR #1303, opened 2022-12-07
feat: add config location to create-definition script

andipaetzold:


packages/contentful--app-scripts/lib/create-app-definition/build-app-definition-settings.js
```diff
@@ -22,9 +22,10 @@ NOTE: This will create an app definition in your Contentful organization.
       message: `Select where your app can be rendered:`,
       type: 'checkbox',
       choices: [
+        { name: 'App configuration screen (app-config) ', value: 'app-config' },
```
loweisz, line 25: We are adding this location by default here: https://github.com/contentful/create-contentful-app/pull/1303/files#diff-5a37f496563485efc42a96165ba34b4e3ab71c809cd0bfb4c42fb47222ac7399L77

So I would also remove it from there

andipaetzold: Oh, my bad. Any reason for adding that by default? There are many apps without config screen.

loweisz: I can't think of a good reason. Maybe check the history? Otherwise I would say it's okay to remove as we already add the dialog location automatically to prevent an error

loweisz (approved):
lgtm

PR #1310, opened 2022-12-08
feat: [EXT-4147] make cca more interactive by default

loweisz:
Adding two new options when providing no example or template. 

Also adding some comments to inform more what currently is going on

Screen Shot 2022-12-13 at 09 38 05

Screen Shot 2022-12-13 at 09 38 34
Screen Shot 2022-12-13 at 09 38 15

Screen Shot 2022-12-13 at 09 38 45


packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -4,29 +4,102 @@ import degit from 'degit';
 import rimraf from 'rimraf';
 
 import { CLIOptions, ContentfulExample } from './types';
-import { highlight, warn } from './logger';
+import { highlight, success, warn, wrapInBlanks } from './logger';
 import { rmIfExists } from './utils';
+// @ts-expect-error no types available
```
ivo-toby, line 9: Is it possible to use `@types/node-fetch` ?

packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -4,29 +4,102 @@ import degit from 'degit';
...
+    ]);
+
+    template = example
+  }
+
+  // return the selected template
+  return selectTemplate(template)
+}
 
 const EXAMPLES_PATH = 'contentful/apps/examples/';
```
ivo-toby, line 72: nit: would be nice to have var's in the beginning of the file

ivo-toby (changes requested):

packages/contentful--create-contentful-app/src/template.ts
loweisz: Weirdly when I tried it the first time it didn't work, but know it seems to work 🤔

packages/contentful--create-contentful-app/src/template.ts
loweisz: 👍

packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -4,29 +4,101 @@ import degit from 'degit';
...
+  ]);
+
+  // if the user chose to use an empty template, ask which language they prefer
+  if (starter === "empty") {
+    const { language } = await inquirer.prompt([
+      {
+        name: 'language',
+        message: 'Do you prefer Typescript or Javascript',
+        type: 'list',
+        choices: ['javascript', 'typescript'],
```
colomolo, line 51: I would switch them places, since typescript is a default choice, so should be first.

packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -4,29 +4,101 @@ import degit from 'degit';
...
+async function promptExampleSelection(): Promise<string> {
+  let template = 'typescript'
+
+  // ask user whether to start with an empty template or use an example template
+  const { starter } = await inquirer.prompt([
+    {
+      name: 'starter',
+      message: 'Do you want to start from an example template or use the empty template?',
+      type: 'list',
+      choices: ['empty', 'template'],
```
colomolo, line 39: ...and here we give choice of empty (what?) or template (empty template or example template?). Feels like we need to change choices to empty and example. And I would use 'blank' instead of 'empty'.

colomolo (changes requested):

packages/contentful--create-contentful-app/src/template.ts
loweisz: Makes sense, I've changed it to "example" and "blank"

packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -4,29 +4,101 @@ import degit from 'degit';
...
+      name: 'starter',
+      message: 'Do you want to start with an example template or use one of the blank templates?',
+      type: 'list',
+      choices: ['blank', 'example'],
+      default: 'blank',
+    },
+  ]);
+
+  // if the user chose to use an empty template, ask which language they prefer
+  if (starter === "empty") {
```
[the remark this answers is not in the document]
loweisz: Good catch, fixed 👍

colomolo (approved):
lgtm

ivo-toby (approved):
👌

packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -4,29 +4,101 @@ import degit from 'degit';
...
+    // get available templates from examples
+    const availableTemplates = await getGithubFolderNames("contentful", "apps", "examples")
+    
+    // ask user to select a template from the available examples
+    const { example } = await inquirer.prompt([
+      {
+        name: 'example',
+        message: 'Select a template',
+        type: 'list',
+        choices: availableTemplates,
```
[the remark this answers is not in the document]
loweisz: Good idea, I think for the example templates that would make sense but for the rather small templates, this would be maybe a bit too much. Let's keep it in mind and see how the adoption is going

PR #1335, opened 2022-12-23
fix: improve capitalization

andipaetzold:
- Ran prettier, so there are more changes than you might expect
- changed "(Using npm)" to "(using npm)" as we are still within the same sentence
- fixed capitalization of TypeScript and JavaScript

Open question: Is it fine to keep "blank" and "example" lowercase?


colomolo:
>  Is it fine to keep "blank" and "example" lowercase?

I think they should be capitalized in options.

Something is not right:
'Do you want to start with an example template or use one of the blank templates?'.

Should be:
'Do you want to start with a blank template or use one of the example templates?'.

ivo-toby:
>  Is it fine to keep "blank" and "example" lowercase?

IMHO it is, they are not names

ivo-toby (approved):
great-job

packages/contentful--create-contentful-app/src/template.ts
```diff
@@ -22,41 +22,44 @@ async function getGithubFolderNames(username: string, repo: string, path: string
...
 async function promptExampleSelection(): Promise<string> {
-  let template = 'typescript'
+  let template = 'typescript';
 
   // ask user whether to start with a blank template or use an example template
   const { starter } = await inquirer.prompt([
     {
       name: 'starter',
-      message: 'Do you want to start with an example template or use one of the blank templates?',
+      message: 'Do you want to start with an example or use one of our blank templates?',
```
colomolo, line 37: Now it feels like there are multiple blank templates and only one example. Maybe something like this?
> Do you want to start with a blank template or use one of our examples?

colomolo: And this also aligns nicely with choices:
1. blank
2. example

colomolo (approved):

PR #1344, opened 2022-12-27
feat: [EXT-4148] Suggest to create an app definition

colomolo:
Suggest creating an App definition right from CLI after bootstrapping an app code.

### DIFF
--- a/packages/contentful--app-scripts/lib/create-app-definition/build-app-definition-settings.js
+++ b/packages/contentful--app-scripts/lib/create-app-definition/build-app-definition-settings.js
@@ -22,12 +22,12 @@ NOTE: This will create an app definition in your Contentful organization.
       message: `Select where your app can be rendered:`,
       type: 'checkbox',
       choices: [
-        { name: 'App configuration screen (app-config) ', value: 'app-config' },
-        { name: 'Entry field (entry-field) ', value: 'entry-field' },
-        { name: 'Entry sidebar (entry-sidebar) ', value: 'entry-sidebar' },
-        { name: 'Entry editor (entry-editor)', value: 'entry-editor' },
-        { name: 'Page (page) ', value: 'page' },
-        { name: 'Home (home) ', value: 'home' },
+        { name: 'App configuration screen ', value: 'app-config' },
+        { name: 'Entry field', value: 'entry-field' },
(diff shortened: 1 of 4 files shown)


packages/contentful--app-scripts/lib/create-app-definition/build-app-definition-settings.js
```diff
@@ -22,12 +22,12 @@ NOTE: This will create an app definition in your Contentful organization.
       message: `Select where your app can be rendered:`,
       type: 'checkbox',
       choices: [
-        { name: 'App configuration screen (app-config) ', value: 'app-config' },
-        { name: 'Entry field (entry-field) ', value: 'entry-field' },
-        { name: 'Entry sidebar (entry-sidebar) ', value: 'entry-sidebar' },
-        { name: 'Entry editor (entry-editor)', value: 'entry-editor' },
-        { name: 'Page (page) ', value: 'page' },
-        { name: 'Home (home) ', value: 'home' },
+        { name: 'App configuration screen ', value: 'app-config' },
```
andipaetzold, line 25: Why are your removing the internal names?

colomolo: Because we also removed them in web app. We also talked about it briefly with Lorenz and decided that we will remove them here.

packages/contentful--create-contentful-app/src/index.ts
```diff
@@ -22,7 +23,16 @@ ${success('Success!')} Created a new Contentful app in ${highlight(tildify(folde
 
   wrapInBlanks(highlight('---- Next Steps'))
 
-  console.log(`Now kick it off by running
+  console.log(`Now create an app definition for your app by running
+
+    ${code(`cd ${tildify(folder)}`)}
+    ${code(useYarn ? 'yarn create-app-definition' : 'npm run create-app-definition')}
+
+    or you can create it manually in web app:
```
andipaetzold, line 31: It's now two "or"s chained. Maybe we can improve the wording a bit to avoid that as it doesn't read smooth

andipaetzold (approved):

packages/contentful--create-contentful-app/src/index.ts
colomolo: Good point. Fixed!

PR #1354, opened 2023-01-04
feat: [] Sketch analytics client

colomolo:


packages/contentful--app-scripts/utils/analytics.js
```diff
@@ -0,0 +1,28 @@
...
+ * @param {string} properties.command triggered command e.g create-app-definition, upload, etc.
+ * @param {boolean} properties.ci value if --ci flag has been set
+ * @returns
+ */
+function track(properties) {
+  if (process.env.DISABLE_ANALYTICS) {
+    return;
+  }
+
+  const client = new Analytics(SEGMENT_WRITE_KEY);
```
colomolo, line 18: Why do we want to create new client every time we execute `track`?

loweisz: With this being a CLI script, it will anyway create a new client for every run, even if we wrap it in another class and instantiate on every call.

loweisz:
Reopened it https://github.com/contentful/create-contentful-app/pull/1356

packages/contentful--app-scripts/utils/analytics.js
```diff
@@ -0,0 +1,33 @@
...
+
+  try {
+    client.track({
+      event: 'app-cli-app-scripts',
+      properties,
+      anonymousId: 'anonymous',
```
colomolo, line 24: I'm not entirely sure, but I think we should not define this id here. This way we'll have a single user for all created apps. We should rather generate unique id every time or remove this field(?):
https://segment.com/docs/connections/sources/catalog/libraries/server/node/#track

image

As I understood it attaches anonymous id automatically:
image

loweisz: Interesting, I think we need to provide at least one of them. So I think I will just create a random ID every call.

PR #1356, opened 2023-01-09
feat: [EXT-4167][EXT-4168][EXT-4169] Create app analytics client and use it in CCA

loweisz:
Creates the tracking client for cca and app-scripts. 

Also uses the tracking function in the template creation command.


packages/contentful--create-contentful-app/src/index.ts
```diff
@@ -106,7 +108,13 @@ async function initProject(appName: string, options: CLIOptions) {
 
     console.log(`Creating a Contentful app in ${highlight(tildify(fullAppFolder))}.`);
 
-    await cloneTemplateIn(fullAppFolder, normalizedOptions);
+    const isInteractive = !normalizedOptions.example && !normalizedOptions.source && !normalizedOptions.javascript && !normalizedOptions.typescript
+
+    const templateSource = await getTemplateSource(options)
+
+    track({ template: templateSource, manager: normalizedOptions.npm ? 'npm' : 'yarn', interactive: isInteractive})
```
colomolo, line 115: Do we want to also track app's (folder) name?

packages/contentful--create-contentful-app/src/index.ts
```diff
@@ -119,6 +127,8 @@ async function initProject(appName: string, options: CLIOptions) {
       await exec('npm', ['install', '--no-audit', '--no-fund'], { cwd: fullAppFolder });
     }
 
+    
+
```
colomolo, line 131: ✂️

packages/contentful--create-contentful-app/src/getTemplateSource.ts
```diff
@@ -0,0 +1,108 @@
...
+}
+
+function selectTemplate(template: string) {
+  wrapInBlanks(highlight(`---- Cloning the ${chalk.cyan(template)} template...`));
+  return EXAMPLES_PATH + template;
+}
+
+
+async function makeContentfulExampleSource(options: CLIOptions): Promise<string> {
+  if (options.example) {
```
colomolo, line 85: Maybe refactor all this to `switch`?

packages/contentful--create-contentful-app/src/getTemplateSource.ts
```diff
@@ -0,0 +1,108 @@
...
+import { highlight, warn, wrapInBlanks } from './logger';
+
+const EXAMPLES_PATH = 'contentful/apps/examples/';
+
+function isContentfulTemplate(url: string) {
+  return Object.values(ContentfulExample).some((t) => url.includes(EXAMPLES_PATH + t));
+}
+
+async function getGithubFolderNames(username: string, repo: string, path: string) {
+  const url = `https://api.github.com/repos/${username}/${repo}/contents/${path}`;
```
colomolo, line 16: Maybe just hardcode the url here? I think it won't change in the nearest observable future.

packages/contentful--app-scripts/utils/analytics.js
```diff
@@ -0,0 +1,33 @@
+const Analytics = require('analytics-node');
```
vikaskumr, line 1: May be use `import`
```suggestion
import Analytics from 'analytics-node';
```

packages/contentful--create-contentful-app/src/index.ts
loweisz: To keep the data as anonymous as possible, I would not do that.

packages/contentful--create-contentful-app/src/getTemplateSource.ts
loweisz: This code was just copy-pasted, but I can change it 👍

packages/contentful--create-contentful-app/src/getTemplateSource.ts
loweisz: Same as above, was just copy-paste but I can change it

packages/contentful--app-scripts/utils/analytics.js
loweisz: This project doesn't have a build process, therefore we can not use modules and need to use require

packages/contentful--create-contentful-app/src/getTemplateSource.ts
loweisz: actually, we are checking here for the existence of different object properties, don't think a switch statement would make much sense here 🤔 Or how would you do that?

colomolo: Ah, yeah, sorry, my bad.

colomolo (approved):
Yeah, looks good to me.

PR #1362, opened 2023-01-11
docs: [] Add section on data gathering to readmes

colomolo:
Add info on tracking for CCA and app-scripts packages.


martin3walker (approved):
LGTM

PR #1377, opened 2023-01-19
feat: [] Separate templates and examples

colomolo:
Adding templates to the list. Currently templates are: Javascript, Typescript, Vue, Next.js, Vite-React.

### DIFF
--- a/packages/contentful--create-contentful-app/src/getTemplateSource.ts
+++ b/packages/contentful--create-contentful-app/src/getTemplateSource.ts
@@ -49,9 +49,9 @@ async function promptExampleSelection(): Promise<string> {
         choices: [
           { name: 'TypeScript', value: 'typescript' },
           { name: 'JavaScript', value: 'javascript' },
-          { name: 'Vue', value: 'vue' },
-          { name: 'Vite + React', value: 'vite-react' },
           { name: 'Next.js', value: 'nextjs' },
+          { name: 'React + Vite', value: 'vite-react' },
+          { name: 'Vue', value: 'vue' },
         ],
         default: 'typescript',
       },
--- a/packages/contentful--create-contentful-app/src/getTemplateSource.ts
+++ b/packages/contentful--create-contentful-app/src/getTemplateSource.ts
@@ -1,55 +1,57 @@
-
 import fetch from 'node-fetch';
 import inquirer from 'inquirer';
 import chalk from 'chalk';
 import { CLIOptions, ContentfulExample } from './types';
 import { highlight, warn, wrapInBlanks } from './logger';
+import { TEMPLATES } from './constants';
 
+const CONTENTFUL_APPS_EXAMPLE_FOLDER =
+  'https://api.github.com/repos/contentful/apps/contents/examples';
 const EXAMPLES_PATH = 'contentful/apps/examples/';
 
 function isContentfulTemplate(url: string) {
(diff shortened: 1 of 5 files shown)


packages/contentful--create-contentful-app/src/getTemplateSource.ts
```diff
@@ -1,55 +1,57 @@
...
-  if (starter === 'blank') {
+  // if the user chose to use a template, ask which language they prefer
+  if (starter === 'template') {
     const { language } = await inquirer.prompt([
       {
         name: 'language',
-        message: 'Do you prefer TypeScript or JavaScript',
+        message: 'Pick a template',
         type: 'list',
         choices: [
```
andipaetzold, line 49: Should we sort by alphabet?

packages/contentful--create-contentful-app/src/getTemplateSource.ts
```diff
@@ -1,55 +1,57 @@
...
           { name: 'TypeScript', value: 'typescript' },
           { name: 'JavaScript', value: 'javascript' },
+          { name: 'Vue', value: 'vue' },
+          { name: 'Vite + React', value: 'vite-react' },
```
andipaetzold, line 53: I'd swap React and Vite because React is more important here. Or is "Vite" more relevant for the user? Yes, the folder name is not ideal here :/

andipaetzold (changes requested):
Two questions

packages/contentful--create-contentful-app/src/getTemplateSource.ts
colomolo: I would rather group them, like:
- TypeScript
- JavaScript
- Next.js
- React + Vite
- Vue

andipaetzold: So, alphabetical with TS and JS as exception. Okay 👍

colomolo: If sorted alphabetically, Javascript will end up too far away from Typescript, which is not convenient, imo.

packages/contentful--create-contentful-app/src/utils.ts
```diff
@@ -77,11 +77,13 @@ export function normalizeOptions(options: CLIOptions): CLIOptions {
   }
 
   if (currentMutuallyExclusiveOptions.length > 1) {
-    warn(
-      `Options ${highlight('--source')}, ${highlight('--example')}, ${highlight(
-        '--typescript'
-      )} and ${highlight('--javascript')} are mutually exclusive, using ${choice(fallbackOption)}.`
-    );
+    const paramsString = `Options ${currentMutuallyExclusiveOptions
```
andipaetzold, line 80: nit: You can use Intl to make this more readable

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat/format

```
new Intl.ListFormat('en', { style: 'long', type: 'conjunction' }).format(currentMutuallyExclusiveOptions);
```
^^ misses `--` prefix and highlighting

andipaetzold (approved):

packages/contentful--create-contentful-app/src/utils.ts
colomolo: Oh, nice, I completely forgot about this stuff.

PR #1475, opened 2023-04-06
feat: [EXT-4404] Check for app actions in manifest and send them on bundle upload

colomolo:


packages/contentful--app-scripts/lib/utils.js
```diff
@@ -59,9 +62,43 @@ const selectFromList = async (list, message, cachedOptionEnvVar) => {
   }
 };
 
+function getActionsManifest() {
+  const isManifestExists = fs.existsSync(DEFAULT_MANIFEST_PATH);
+
+  if (!isManifestExists) {
```
martin3walker, line 68: Would it make sense to provide an error here that says "Manifest file required at './contentful-app-manifest.json'" just in case someone changes the name for some reason?

martin3walker (commented):
Minor comment, but looks good!

packages/contentful--app-scripts/lib/utils.js
colomolo: But it's not required afaik. So we can't really check if some random file is a manifest or not.

martin3walker: Ah right, fair point

martin3walker (approved):
LGTM

PR #1492, opened 2023-04-20
fix: clone app action templates correctly

martin3walker:


packages/contentful--create-contentful-app/src/includeAppAction.ts
```diff
@@ -1,62 +1,61 @@
...
+import degit from 'degit';
+import { success, highlight } from './logger';
 
-export function cloneAppAction(templateIsTypescript: boolean, destination: string) {
-  const actionPath = resolve(
-    templateIsTypescript
-      ? 'src/app-actions/typescript/index.ts'
-      : 'src/app-actions/javascript/index.js'
-  );
+export async function cloneAppAction(templateIsTypescript: boolean, destination: string) {
```
colomolo, line 8: Strange that non-primary boolean param goes first. Should be second and non-mandatory, I think (with fallback to JS IMO).

packages/contentful--create-contentful-app/src/includeAppAction.ts
```diff
@@ -1,62 +1,61 @@
...
+  try {
+    console.log(highlight('Cloning hosted app action.'));
+    // Clone the app actions template to the created directory under the folder 'actions'
+    const templateSource = `contentful/apps/examples/hosted-app-action-templates/${
```
colomolo, line 12:
```suggestion
    const templateSource = path.join('contentful/apps/examples/hosted-app-action-templates', templateIsTypescript ? 'typescript' : 'javascript');
```

martin3walker: Is this just for readability, or is there something else that it does?

packages/contentful--create-contentful-app/src/includeAppAction.ts
martin3walker: I can switch the order, but I think it needs to be mandatory. If we're not explicit we run the risk of cloning js files into a typescript repo, which will cause problems for users. Wdyt?

colomolo: Maybe it's better to switch it from boolean to `'typescript' | 'javascript'` type to support other templates in the future?

colomolo: > If we're not explicit we run the risk of cloning js files into a typescript repo, which will cause problems for users.

Yes, makes sense

colomolo (approved):

packages/contentful--create-contentful-app/src/includeAppAction.ts
martin3walker: Nice, thanks!

PR #1523, opened 2023-05-10
allowNetworks argument added in app action manifest[EXT-4509]

vikaskumr:
Adding new endpoint property `allowNetworks` in the action manifest file.

User should provide the list of allowed networks in action manifest file.

By default the below networks will always have access
`api.contentful.com, upload.contentful.com, 127.0.0.1`

### DIFF
--- a/packages/contentful--app-scripts/lib/utills.test.js
+++ b/packages/contentful--app-scripts/lib/utills.test.js
@@ -45,14 +45,9 @@ describe('removeProtocolFromUrl', () => {
     assert.strictEqual(result, 'example.com');
   });
 
-  it('returns undefined for an invalid URL', () => {
-    const result = removeProtocolFromUrl('not a url');
-    assert.strictEqual(result, undefined);
-  });
-
   it('returns valid ipv6 address', () => {
-    const result = removeProtocolFromUrl('2001:0db8:85a3:0000:0000:8a2e:0370:7334');
-    assert.strictEqual(result, '[2001:db8:85a3::8a2e:370:7334]');
+    const result = removeProtocolFromUrl('http://[2001:0db8:85a3:0000:0000:8a2e:0370]:7334');
+    assert.strictEqual(result, '[2001:0db8:85a3:0000:0000:8a2e:0370]:7334');
   });
 
   it('returns valid domain', () => {
--- a/packages/contentful--app-scripts/lib/utills.test.js
+++ b/packages/contentful--app-scripts/lib/utills.test.js
@@ -44,4 +44,19 @@ describe('removeProtocolFromUrl', () => {
     const result = removeProtocolFromUrl('not a url');
     assert.strictEqual(result, undefined);
   });
+
+  it('returns valid ipv6 address', () => {
+    const result = removeProtocolFromUrl('2001:0db8:85a3:0000:0000:8a2e:0370:7334');
+    assert.strictEqual(result, '[2001:db8:85a3::8a2e:370:7334]');
+  });
+
+  it('returns valid domain', () => {
+    const result = removeProtocolFromUrl('example.com');
+    assert.strictEqual(result, 'example.com');
+  });
+
+  it('returns valid domain with port', () => {
+    const result = removeProtocolFromUrl('example.com:40');
+    assert.strictEqual(result, 'example.com:40');
+  });
 });
--- a/packages/contentful--app-scripts/lib/utills.test.js
+++ b/packages/contentful--app-scripts/lib/utills.test.js
@@ -32,6 +32,11 @@ describe('isValidIpAddress', () => {
     const result = isValidIpAddress('not an ip address');
     assert.strictEqual(result, false);
   });
+
+  it('returns true for an valid ipv6 address with port', () => {
+    const result = isValidIpAddress('[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:443');
+    assert.strictEqual(result, true);
+  });
 });
 
 describe('removeProtocolFromUrl', () => {
--- a/packages/contentful--app-scripts/lib/utills.test.js
+++ b/packages/contentful--app-scripts/lib/utills.test.js
@@ -1,0 +1,47 @@
+const assert = require('assert');
+
+const { isValidIpAddress, removeProtocolFromUrl } = require('./utils');
+
+describe('isValidIpAddress', () => {
+  it('returns true for a valid IP address', () => {
+    const result = isValidIpAddress('192.168.0.1');
+    assert.strictEqual(result, true);
+  });
+
+  it('returns true for a valid domain', () => {
+    const result = isValidIpAddress('http://google.com');
+    assert.strictEqual(result, true);
+  });
+
+  it('returns true for a valid ipv6 address', () => {
+    const result = isValidIpAddress('2001:0db8:85a3:0000:0000:8a2e:0370:7334');
+    assert.strictEqual(result, true);
+  });
+
+  it('returns true for a valid domain with port', () => {
+    const result = isValidIpAddress('http://google.com:4000');
+    assert.strictEqual(result, true);
+  });
+
+  it('returns true for a valid IP address with port', () => {
+    const result = isValidIpAddress('192.168.0.1:2000');
+    assert.strictEqual(result, true);
+  });
+
+  it('returns false for an invalid IP address', () => {
+    const result = isValidIpAddress('not an ip address');
+    assert.strictEqual(result, false);
+  });
+});
+
+describe('removeProtocolFromUrl', () => {
+  it('returns the host of a URL without a protocol', () => {
+    const result = removeProtocolFromUrl('http://example.com');
+    assert.strictEqual(result, 'example.com');
+  });
+
+  it('returns undefined for an invalid URL', () => {
+    const result = removeProtocolFromUrl('not a url');
+    assert.strictEqual(result, undefined);
+  });
+});
--- a/packages/contentful--app-scripts/lib/utills.test.js
+++ b/packages/contentful--app-scripts/lib/utills.test.js
@@ -9,7 +9,7 @@ describe('isValidIpAddress', () => {
   });
 
   it('returns true for a valid domain', () => {
-    const result = isValidIpAddress('http://google.com');
+    const result = isValidIpAddress('google.com');
     assert.strictEqual(result, true);
   });
 
@@ -19,7 +19,7 @@ describe('isValidIpAddress', () => {
   });
 
   it('returns true for a valid domain with port', () => {
-    const result = isValidIpAddress('http://google.com:4000');
+    const result = isValidIpAddress('google.com:4000');
     assert.strictEqual(result, true);
   });
 
--- a/packages/contentful--app-scripts/lib/utills.test.js
+++ b/packages/contentful--app-scripts/lib/utills.test.js
@@ -1,62 +1,62 @@
 const assert = require('assert');
 
-const { isValidIpAddress, removeProtocolFromUrl } = require('./utils');
+const { isValidNetwork, stripProtocol } = require('./utils');
 
 describe('isValidIpAddress', () => {
   it('returns true for a valid IP address', () => {
-    const result = isValidIpAddress('192.168.0.1');
+    const result = isValidNetwork('192.168.0.1');
     assert.strictEqual(result, true);
   });
 
   it('returns true for a valid domain', () => {
-    const result = isValidIpAddress('google.com');
+    const result = isValidNetwork('google.com');
     assert.strictEqual(result, true);
   });
 
   it('returns true for a valid ipv6 address', () => {
-    const result = isValidIpAddress('2001:0db8:85a3:0000:0000:8a2e:0370:7334');
+    const result = isValidNetwork('2001:0db8:85a3:0000:0000:8a2e:0370:7334');
     assert.strictEqual(result, true);
   });
 
   it('returns true for a valid domain with port', () => {
-    const result = isValidIpAddress('google.com:4000');
+    const result = isValidNetwork('google.com:4000');
     assert.strictEqual(result, true);
   });
 
   it('returns true for a valid IP address with port', () => {
-    const result = isValidIpAddress('192.168.0.1:2000');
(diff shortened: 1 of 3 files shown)


packages/contentful--app-scripts/lib/utils.js
```diff
@@ -13,6 +14,22 @@ const throwValidationException = (subject, message, details) => {
   throw new TypeError(message);
 };
 
+const validateIpAddress = (ipAddress) => {
```
kdamball, line 17: can we add some unit tests for these? It's a big regex and brittle. Also the name can be `isValidIpAddress` to hint that we're returning boolean

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -81,7 +98,33 @@ function getActionsManifest() {
   ----------------------------`);
     console.log('');
 
-    return manifest.actions.map((action) => ({ parameters: [], ...action })); // adding required parameters
+    const actions = manifest.actions.map((action) => {
+      const allowedNetworks = Array.isArray(action.allowedNetworks)
+        ? [...action.allowedNetworks, 'api.contentful.com', 'upload.contentful.com', '127.0.0.1']
+        : ['api.contentful.com', 'upload.contentful.com', '127.0.0.1'];
+
+      const validAllowedNetworks = allowedNetworks
```
kdamball, line 106: You can simplify this with `hasInvalidNetwork = allowedNetworks.map().some(netWork => !isValid())`

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -81,7 +98,33 @@ function getActionsManifest() {
   ----------------------------`);
     console.log('');
 
-    return manifest.actions.map((action) => ({ parameters: [], ...action })); // adding required parameters
+    const actions = manifest.actions.map((action) => {
+      const allowedNetworks = Array.isArray(action.allowedNetworks)
+        ? [...action.allowedNetworks, 'api.contentful.com', 'upload.contentful.com', '127.0.0.1']
```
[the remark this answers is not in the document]
vikaskumr: ohh ok I thought we need to add the default values here :)

packages/contentful--app-scripts/lib/utils.js
kdamball: Dont forget IPv6

andipaetzold: Isn't there a library for that? 😶 I wouldn't reinvent the wheel for such a big regex.

vikaskumr: I will look for the library.

vikaskumr: I searched for library, but unfortunately I couldn't find all the things which I need in one library

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -81,7 +108,31 @@ function getActionsManifest() {
...
+          )} Invalid IP address found in the allowedNetworks array for action "${action.name}".`
+        );
+        // eslint-disable-next-line no-process-exit
+        process.exit(1);
+      }
+
+      return {
+        parameters: [],
+        ...action,
+        allowedNetworks: allowedNetworks,
```
colomolo, line 131:
```suggestion
        allowedNetworks,
```

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -13,6 +14,32 @@ const throwValidationException = (subject, message, details) => {
   throw new TypeError(message);
 };
 
+const isValidIpAddress = (address) => {
+  const addressRegex =
+    /^(?:(?:https?|ftp):\/\/)?(?:localhost|(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,6}|(?:\d{1,3}\.){3}\d{1,3}|(\[(?:[A-Fa-f0-9]{1,4}:){7}[A-Fa-f0-9]{1,4}\]|(?:[A-Fa-f0-9]{1,4}:){7}[A-Fa-f0-9]{1,4}))(?::\d{1,5})?$/;
```
ivo-toby, line 19: looks like its matching for protocols, which you remove in ln 115, so why still match for those?

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -13,6 +14,32 @@ const throwValidationException = (subject, message, details) => {
   throw new TypeError(message);
 };
 
+const isValidIpAddress = (address) => {
```
ivo-toby, line 17: Nit: technically this method name is incorrect, is also checks fqdn's, not only ip's

packages/contentful--app-scripts/lib/utills.test.js
```diff
@@ -0,0 +1,62 @@
...
+const { isValidIpAddress, removeProtocolFromUrl } = require('./utils');
+
+describe('isValidIpAddress', () => {
+  it('returns true for a valid IP address', () => {
+    const result = isValidIpAddress('192.168.0.1');
+    assert.strictEqual(result, true);
+  });
+
+  it('returns true for a valid domain', () => {
+    const result = isValidIpAddress('http://google.com');
```
ivo-toby, line 12: This should have failed, we don't want the protocol in there

packages/contentful--app-scripts/lib/utills.test.js
```diff
@@ -0,0 +1,62 @@
...
+    assert.strictEqual(result, true);
+  });
+
+  it('returns true for a valid ipv6 address', () => {
+    const result = isValidIpAddress('2001:0db8:85a3:0000:0000:8a2e:0370:7334');
+    assert.strictEqual(result, true);
+  });
+
+  it('returns true for a valid domain with port', () => {
+    const result = isValidIpAddress('http://google.com:4000');
```
ivo-toby, line 22: Same

packages/contentful--app-scripts/lib/utills.test.js
ivo-toby, line 1: Maybe worth testing for a valid ipv6 address with port, ie: `[2001:0db8:85a3:0000:0000:8a2e:0370:7334]:443`

ivo-toby (changes requested):

packages/contentful--app-scripts/lib/utils.js
vikaskumr: Users can still have input the domains with protocol, according to them it is correct domain but for our case we don't need it so that is why I am removing protocol from this `removeProtocolFromUrl` function.

According to the ticket I 🤔 we can accept it but we can remove it internally and then I can send it to backend

packages/contentful--app-scripts/lib/utills.test.js
vikaskumr: I am removing removing protocol in `removeProtocolFromUrl` and since we are accepting the domain with protocol, it is correct unless it is otherwise.

@kdamball wdyt?

packages/contentful--app-scripts/lib/utils.js
ivo-toby: I see, but you are removing protocol with removeProtocolFromUrl before checking with this method right? 
It's not that big of a deal, it's just that it's checking for things that are not relevant.

packages/contentful--app-scripts/lib/utills.test.js
ivo-toby: Ok, I understand, but that implies things to be done in other methods that you are not testing here.

packages/contentful--app-scripts/lib/utils.js
vikaskumr: Yes, make sense

packages/contentful--app-scripts/package.json
```diff
@@ -27,7 +27,7 @@
     "prettier": "prettier **/*.js --write --ignore-path .gitignore",
     "lint": "eslint ./lib",
     "lint:fix": "npm run lint -- --fix",
-    "test": "mocha ./lib/**/*.test.js ./lib/*.test.js ./utils/**/*.test.js",
```
kdamball, line 30: Why is this deleted?

vikaskumr: by mistake 😔

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -13,6 +13,25 @@ const throwValidationException = (subject, message, details) => {
...
+  const addressRegex =
+    /^(?:localhost|(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,6}|(?:\d{1,3}\.){3}\d{1,3}|(\[(?:[A-Fa-f0-9]{1,4}:){7}[A-Fa-f0-9]{1,4}\]|(?:[A-Fa-f0-9]{1,4}:){7}[A-Fa-f0-9]{1,4}))(?::\d{1,5})?$/;
+  return addressRegex.test(address);
+};
+
+const removeProtocolFromUrl = (url) => {
+  try {
+    const protocolRemovedUrl = url.replace(/^https?:\/\//, '');
+
+    const host = protocolRemovedUrl.split('/')[0];
```
colomolo, line 26:
```suggestion
    return protocolRemovedUrl.split('/')[0];
```

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -13,6 +13,25 @@ const throwValidationException = (subject, message, details) => {
   throw new TypeError(message);
 };
 
+const isValidNetwork = (address) => {
+  const addressRegex =
+    /^(?:localhost|(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,6}|(?:\d{1,3}\.){3}\d{1,3}|(\[(?:[A-Fa-f0-9]{1,4}:){7}[A-Fa-f0-9]{1,4}\]|(?:[A-Fa-f0-9]{1,4}:){7}[A-Fa-f0-9]{1,4}))(?::\d{1,5})?$/;
+  return addressRegex.test(address);
+};
+
+const removeProtocolFromUrl = (url) => {
```
colomolo, line 22: Maybe name it just `stripProtocol`?

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -81,7 +100,31 @@ function getActionsManifest() {
   ----------------------------`);
     console.log('');
 
-    return manifest.actions.map((action) => ({ parameters: [], ...action })); // adding required parameters
+    const actions = manifest.actions.map((action) => {
+      const allowedNetworks = Array.isArray(action.allowedNetworks) ? action.allowedNetworks : [];
+
+      const hasInvalidNetworks = allowedNetworks
+        .map(removeProtocolFromUrl)
+        .some((netWork) => !isValidNetwork(netWork));
```
colomolo, line 108: I would replace `some` with `find` to be able to get exact invalid network and show it in error message below. Same computing price, but we can give a bit more info to the user that way.

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -81,7 +100,31 @@ function getActionsManifest() {
   ----------------------------`);
     console.log('');
 
-    return manifest.actions.map((action) => ({ parameters: [], ...action })); // adding required parameters
+    const actions = manifest.actions.map((action) => {
+      const allowedNetworks = Array.isArray(action.allowedNetworks) ? action.allowedNetworks : [];
```
colomolo, line 104: Can map here instead of doing it twice below.
```suggestion
      const allowedNetworks = Array.isArray(action.allowedNetworks) ? action.allowedNetworks.map(removeProtocolFromUrl) : [];
```

packages/contentful--app-scripts/lib/utills.test.js
```diff
@@ -0,0 +1,62 @@
...
+    assert.strictEqual(result, 'example.com');
+  });
+
+  it('returns valid ipv6 address', () => {
+    const result = removeProtocolFromUrl('http://[2001:0db8:85a3:0000:0000:8a2e:0370]:7334');
+    assert.strictEqual(result, '[2001:0db8:85a3:0000:0000:8a2e:0370]:7334');
+  });
+
+  it('returns valid domain', () => {
+    const result = removeProtocolFromUrl('example.com');
```
colomolo, line 54: Should have protocol here, right? Or we can test both cases – with and without a protocol.

packages/contentful--app-scripts/lib/utills.test.js
```diff
@@ -0,0 +1,62 @@
...
+    assert.strictEqual(result, 'example.com');
+  });
+
+  it('returns valid domain with port', () => {
+    const result = removeProtocolFromUrl('example.com:40');
```
colomolo, line 59: Same here

packages/contentful--app-scripts/lib/utills.test.js
colomolo: Ah, nevermind, you test it above.

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -13,6 +13,25 @@ const throwValidationException = (subject, message, details) => {
...
 };
 
+const isValidNetwork = (address) => {
+  const addressRegex =
+    /^(?:localhost|(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?\.)+[a-zA-Z]{2,6}|(?:\d{1,3}\.){3}\d{1,3}|(\[(?:[A-Fa-f0-9]{1,4}:){7}[A-Fa-f0-9]{1,4}\]|(?:[A-Fa-f0-9]{1,4}:){7}[A-Fa-f0-9]{1,4}))(?::\d{1,5})?$/;
+  return addressRegex.test(address);
+};
+
+const removeProtocolFromUrl = (url) => {
+  try {
```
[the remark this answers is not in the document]
vikaskumr: I was previously using URL library but after removing forgot to remove try/catch

kdamball (approved):

PR #1524, opened 2023-05-12
fix: [] fix tracking schema violation

loweisz:
Fixes the type of the argument for the schema. was expecting a string but got a boolean


vikaskumr (approved):
:shipit:

PR #1526, opened 2023-05-12
fix: build script for hosted actions[EXT-4509]

vikaskumr:
Now we have build script for both JavaScript and TypeScript templates for hosted actions.


kdamball (approved):
LGTM - only merge after the other one

PR #1530, opened 2023-05-17
Update app manifest with entry file prop

ivo-toby:


packages/contentful--app-scripts/lib/utils.js
```diff
@@ -112,9 +111,13 @@ function getActionsManifest() {
...
       }
 
+      // EntryFile is not used but we do want to strip it from action
+      // eslint-disable-next-line no-unused-vars
+      const { entryFile: _, ...rest } = action;
+
       return {
         parameters: [],
-        ...action,
+        ...rest,
```
vikaskumr, line 120: may be a better name `actionWithoutEntryFile`

vikaskumr (approved):
Looks Good :shipit:

PR #1535, opened 2023-05-19
feat: support host flag in more upload and activate commands

ivo-toby:


packages/contentful--create-contentful-app/package-lock.json
andipaetzold, line 1: Nothing should change here

packages/contentful--app-scripts/bin/app-scripts
```diff
@@ -29,6 +29,7 @@ async function runCommand(command, options) {
     .option('--token [accessToken]', 'Your content management access token')
     .option('--comment [comment]', 'Optional comment for the created bundle')
     .option('--skip-activation', 'A Boolean flag to skip automatic activation')
+    .option('--host [host]', 'Contentful CMA-endpoint to use')
```
andipaetzold, line 32: We don't specify the endpoint here. We only specify the domain/host.

Also, imo the word "CMA-endpoint" doesn't look pretty 😬

packages/contentful--app-scripts/lib/activate/activate-bundle.test.js
```diff
@@ -10,7 +10,8 @@ const mockedSettings = {
 };
 
 describe('activate-bundle', () => {
-  let activateBundle, clientMock, updateStub;
+  // eslint-disable-next-line no-unused-vars
```
andipaetzold, line 13: Why unused?

andipaetzold (approved):
just nits

packages/contentful--app-scripts/lib/activate/activate-bundle.test.js
ivo-toby: Code was nagging about it, though it's actually used

PR #1631, opened 2023-08-11
feat: [] Prompt all existing options in interactive mode

colomolo:
Prompting `host`, `keep`, and `skipActivation` options while running app-scripts in interactive mode. Nothing will change for non-interactive mode.


packages/contentful--app-scripts/utils/constants.js
```diff
@@ -7,12 +7,15 @@ const DEFAULT_BUNDLES_TO_KEEP = 50;
 const DEFAULT_BUNDLES_TO_FETCH = 1000;
 const MAX_CONCURRENT_DELETION_CALLS = 5;
 
+const CONTENTFUL_API_HOST = 'api.contentful.com';
```
andipaetzold, line 10: Should we previx with `DEFAULT_`?

packages/contentful--app-scripts/lib/upload/build-upload-settings.js
```diff
@@ -22,6 +23,21 @@ async function buildAppUploadSettings(options) {
       default: '',
     });
   }
+  if (!skipActivation) {
+    prompts.push({
+      type: 'confirm',
+      name: 'skipActivation',
+      message: `Skip bundle activation after upload?`,
```
andipaetzold, line 30: 🤔  To me it feels like this question would be easier to understand in reverse:

> Do you want to activate the bundle after upload?

packages/contentful--app-scripts/lib/upload/build-upload-settings.js
```diff
@@ -22,6 +23,21 @@ async function buildAppUploadSettings(options) {
       default: '',
     });
   }
+  if (!skipActivation) {
```
andipaetzold, line 26: We'd need to check for `skipActivation === undefined`, because `skipActivation` is a boolean and we might already have explicitely set it to false.

packages/contentful--app-scripts/lib/upload/build-upload-settings.js
andipaetzold: Does it work now? We set `skipActivation` to true if we want to activate the bundle. I feel like we need to map from the user input to `skipActivation`

colomolo: Oh, yes, good catch. Fixed it now.

andipaetzold (approved):

PR #1634, opened 2023-08-14
feat: [] Convert app-scripts to TS

colomolo:
Switching `app-scripts` packages to TypeScript.


packages/contentful--app-scripts/tsconfig.json
```diff
@@ -0,0 +1,14 @@
...
+  "exclude": ["node_modules", "**/*.test.ts"],
+  "compilerOptions": {
+    "target": "es2016",
+    "module": "commonjs",
+    "esModuleInterop": true,
+    "forceConsistentCasingInFileNames": true,
+    "strict": true,
+    "skipLibCheck": true,
+    "outDir": "./lib",
+    "typeRoots": ["./typings", "./node_modules/@types"]
```
andipaetzold, line 12: Can we try to remove as many options as possible here, so our main config comes from `@tsconfig/node16`? Besides `outDir`, we can probably drop everything.

packages/contentful--app-scripts/package.json
```diff
@@ -61,5 +62,16 @@
     "open": "8.4.2",
     "ora": "5.4.1"
   },
-  "gitHead": "4c3506be3f52c7a8aae17deaa75acefc9a805b42"
+  "gitHead": "4c3506be3f52c7a8aae17deaa75acefc9a805b42",
+  "devDependencies": {
+    "@tsconfig/node16": "^16.1.0",
```
andipaetzold, line 67: I think we are still supporting `node14`, right? So we should use that as tsconfig

packages/contentful--app-scripts/package.json
```diff
@@ -61,5 +62,16 @@
...
+    "@types/adm-zip": "^0.5.0",
+    "@types/analytics-node": "^3.1.11",
+    "@types/inquirer": "8.2.1",
+    "@types/lodash": "^4.14.197",
+    "@types/mocha": "^10.0.1",
+    "@types/proxyquire": "^1.3.28",
+    "@types/sinon": "^10.0.16",
+    "ts-node": "^10.9.1"
```
andipaetzold, line 75: Can we pin those?

packages/contentful--app-scripts/package.json
```diff
@@ -13,7 +13,7 @@
     "node": ">=14",
     "npm": ">=6"
   },
-  "main": "lib/index.js",
+  "main": "lib/src/index.js",
```
andipaetzold, line 16: This often indicates that we compile to much and the compiled test files end up in our `lib` folder. I think we can solve that by being explicit about what to compile in the tsconfig (using `files` or `includes`)

packages/contentful--app-scripts/src/activate/get-activate-args.ts
```diff
@@ -1,16 +1,17 @@
-const chalk = require('chalk');
-const ora = require('ora');
-const { getAppInfo } = require('../get-app-info');
-const { validateArguments } = require('../validate-arguments');
+import chalk from 'chalk';
+import ora from 'ora';
+import { getAppInfo } from '../get-app-info';
+import { validateArguments } from '../validate-arguments';
+import { ActivateOptions, ActivateSettings } from '.';
```
andipaetzold, line 5: By importing from `.` we are creating a circular dependency. It currently works because TypeScript is smart, but I'd avoid doing that.

packages/contentful--app-scripts/src/activate/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  await activateBundle(settings);
+};
+
+const nonInteractive = async (options: ActivateOptions) => {
+  const settings = await getActivateSettingsArgs(options);
+
+  await activateBundle(settings);
+};
+
+export default {
```
andipaetzold, line 34: Can we do named exports here? Or would that be a breaking change?

packages/contentful--app-scripts/src/clean-up/clean-up-bundles.ts
```diff
@@ -57,9 +71,9 @@ async function cleanUpBundles(settings) {
     if (requestedAmount < 1) {
       throw new Error('Requested amount of bundles to fetch must be greater than 0');
     }
-    const getBundles = async (limit, skip) => {
+    const getBundles = async (limit?: number, skip?: number): Promise<AppBundleProps[]> => {
       const result = await fetchAppBundles(limit, skip);
-      const currLength = skip + result.items.length;
+      const currLength = (skip ?? 0) + result.items.length;
```
andipaetzold, line 76: Nit: we can set the default for `skip` already in the function args.

packages/contentful--app-scripts/utils/cache-credential/index.ts
```diff
@@ -1,9 +1,9 @@
-const dotenv = require('dotenv');
-const fs = require('fs');
-const { EOL } = require('os');
+import dotenv from 'dotenv';
+import fs from 'fs';
+import { EOL } from 'os';
 const ignore = require('ignore')();
```
andipaetzold, line 4: Any reason for not changing this import?

packages/contentful--app-scripts/src/utils.ts
```diff
@@ -1,31 +1,44 @@
...
+import chalk from 'chalk';
+import inquirer from 'inquirer';
+import { cacheEnvVars } from '../utils/cache-credential';
+import { Definition } from './definition-api';
+import { Organization } from './organization-api';
 
 const DEFAULT_MANIFEST_PATH = './contentful-app-manifest.json';
 
-const throwValidationException = (subject, message, details) => {
+export interface FunctionAppAction {
```
andipaetzold, line 10: Isn't that hosted app actions?

andipaetzold (changes requested):
Finally typescript ❤️ 

Got a few comments - nothing marjor though

packages/contentful--app-scripts/tsconfig.json
colomolo: i've just copied from create-contentful-app to be consistent, but ok, will clean this up.

packages/contentful--app-scripts/src/activate/index.ts
colomolo: Should work.

packages/contentful--app-scripts/package.json
colomolo: Wdym? Previously it was a `lib` folder with all js files inside, now I moved all to `src` and output is in `lib`. We excluding test files here: https://github.com/contentful/create-contentful-app/pull/1634/files#diff-62dd7532f57cf9b72c5d182426154c8422775ea59455ea5178d2a1a4b39130c9R3

packages/contentful--app-scripts/utils/cache-credential/index.ts
colomolo: missed this one.

packages/contentful--app-scripts/package.json
andipaetzold: The folder structure now looks like this

image

But we normally want it to be like this

image

Probably can easily be fixed by moving the `utils` folder into the `src` folder.

colomolo: Oh nice, I didn't know how to solve that. Thanks!

packages/contentful--app-scripts/src/utils.ts
colomolo: No, this is slightly different and should be aligned with [schema](https://github.com/contentful/extensibility-api/blob/master/lib/entities/app-action/schema.ts), except `entryFile` field.

colomolo:
@kdamball Can you please take a look. I've just merged your changes.

packages/contentful--app-scripts/bin/app-scripts
andipaetzold, line 1: Should we migrate this to TS as well?

packages/contentful--app-scripts/src/create-app-definition/create-app-definition.ts
```diff
@@ -55,7 +56,7 @@ function assertValidArguments(accessToken, appDefinitionSettings) {
   }
 }
 
-async function createAppDefinition(accessToken, appDefinitionSettings = { locations: [] }) {
+export async function createAppDefinition(accessToken: string, appDefinitionSettings: AppDefinitionSettings) {
```
andipaetzold, line 59: We previously had a default value for `appDefinitionSettings`. Now we only have a default value for `appDefinitionSettings.fields`. Could this cause issues for some code paths or is `appDefinitionSettings` always defined`?

packages/contentful--app-scripts/src/upload/create-app-upload.ts
```diff
@@ -0,0 +1,37 @@
...
+import { UploadSettings } from '../types';
+
+async function createAppBundleFromFile(orgId: string, token: string, zip: Buffer, host = '') {
+  const client = createClient({ accessToken: token, host, hostUpload: host.replace(/^api/i, 'upload') });
+  const org = await client.getOrganization(orgId);
+  return await org.createAppUpload(zip);
+}
+
+export async function createAppUpload(settings: UploadSettings) {
+  validateBundle(settings.bundleDirectory || '../types');
```
andipaetzold, line 15: `types`? In the js code it's `.`

andipaetzold (changes requested):
Almost good to go

packages/contentful--app-scripts/src/create-app-definition/create-app-definition.ts
colomolo: Yeah, we ask for them in interactive mode, so they are always present.

packages/contentful--app-scripts/src/analytics.ts
```diff
@@ -7,10 +7,10 @@ const SEGMENT_WRITE_KEY = 'IzCq3j4dQlTAgLdMykRW9oBHQKUy1xMm';
  *
  * @param {object} properties tracking properties
  * @param {string} properties.command triggered command e.g create-app-definition, upload, etc.
- * @param {boolean} properties.ci value if --ci flag has been set
+ * @param {string} properties.ci value if --ci flag has been set
```
andipaetzold, line 10: String?

packages/contentful--app-scripts/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "@tsconfig/node14/tsconfig.json",
+  "exclude": ["node_modules", "**/*.test.ts"],
+  "compilerOptions": {
+    "outDir": "./lib"
```
andipaetzold, line 5: Can we enable `declaration`, so people have types when importing from this package? This would also require a `types` prop in package.json.

andipaetzold (changes requested):

packages/contentful--app-scripts/src/analytics.ts
colomolo: Yes, we pass string here, for example:
https://github.com/contentful/create-contentful-app/pull/1634/files#diff-93c382db7222a985fff313bc92ddd886018770fe7872c55bbf06dce16b6fff52R72

andipaetzold: Feels quite wrong as it's a boolean flag 🤔

colomolo: Schema accepts string, I guess:
https://github.com/contentful/create-contentful-app/pull/1524

colomolo: But we can convert it to string inside of `track`, I believe.

packages/contentful--app-scripts/package.json
andipaetzold, line 1: `"types": "lib/index.d.ts"` is missing

andipaetzold (changes requested):
One change and we are good to go

packages/contentful--app-scripts/package.json
colomolo: Now should be good

PR #1636, opened 2023-08-15
Add --delivery-function option to add the delivery function example to create-contentful-app [MONET-1334]

BasKiers:
The PR includes the following changes
- Adds a new `--delivery-function` option to `create-contentful-app`
- Merges content from `package.json` in delivery function template folder into the project's `package.json`
- Fixes a small bug encountered during testing which prevented the app directory from being removed on an unexpected error

### DIFF
--- a/packages/contentful--create-contentful-app/src/includeAppAction.ts
+++ b/packages/contentful--create-contentful-app/src/includeAppAction.ts
@@ -47,6 +47,7 @@ export async function cloneAppAction(destination: string, templateIsTypescript: 
     }
 
     await mergeJsonIntoFile({
+      source: `${appActionDirectoryPath}/package.json`,
       destination: packageJsonLocation,
       mergeFn: addBuildCommand,
     });
--- a/packages/contentful--create-contentful-app/src/includeAppAction.ts
+++ b/packages/contentful--create-contentful-app/src/includeAppAction.ts
@@ -1,9 +1,16 @@
 import inquirer from 'inquirer';
-import { readFileSync, writeFileSync, existsSync, renameSync } from 'fs';
+import { rename } from 'fs/promises';
 import { resolve, join } from 'path';
 import { CONTENTFUL_APP_MANIFEST } from './constants';
 import degit from 'degit';
-import { success, highlight } from './logger';
+import { highlight } from './logger';
+import { getAddBuildCommandFn } from './utils/package';
+import { exists, mergeJsonIntoFile } from './utils/file';
+
+const addBuildCommand = getAddBuildCommandFn({
+  name: 'build:actions',
+  command: 'node build-actions.js',
+});
 
 export async function cloneAppAction(destination: string, templateIsTypescript: boolean) {
   try {
@@ -11,7 +18,7 @@ export async function cloneAppAction(destination: string, templateIsTypescript: 
     // Clone the app actions template to the created directory under the folder 'actions'
     const templateSource = join(
       'contentful/apps/examples/hosted-app-action-templates',
-      templateIsTypescript ? 'typescript' : 'javascript'
+      templateIsTypescript ? 'typescript' : 'javascript',
     );
 
     const appActionDirectoryPath = resolve(`${destination}/actions`);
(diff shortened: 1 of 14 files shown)


packages/contentful--create-contentful-app/src/includeAppAction.ts
```diff
@@ -20,33 +27,30 @@ export async function cloneAppAction(destination: string, templateIsTypescript:
...
-      `${destination}/${CONTENTFUL_APP_MANIFEST}`
-    );
+    await mergeJsonIntoFile({
+      source: `${appActionDirectoryPath}/${CONTENTFUL_APP_MANIFEST}`,
+      destination: `${destination}/${CONTENTFUL_APP_MANIFEST}`,
+    });
 
     // move the build file from the actions folder to the root folder
-    renameSync(`${appActionDirectoryPath}/build-actions.js`, `${destination}/build-actions.js`);
+    await rename(`${appActionDirectoryPath}/build-actions.js`, `${destination}/build-actions.js`);
```
kdamball, line 36: If we're moving to async, we can do this js file and json merge above in parallel :)

packages/contentful--create-contentful-app/src/includeDeliveryFunction.ts
```diff
@@ -0,0 +1,60 @@
...
+    await d.clone(deliveryFunctionDirectoryPath);
+
+    // merge the manifest from the template folder to the root folder
+    await mergeJsonIntoFile({
+      source: `${deliveryFunctionDirectoryPath}/${CONTENTFUL_APP_MANIFEST}`,
+      destination: `${destination}/${CONTENTFUL_APP_MANIFEST}`,
+    });
+
+    // move the build file from the actions folder to the root folder
+    await rename(
```
kdamball, line 35: Same thing here

packages/contentful--create-contentful-app/src/index.ts
```diff
@@ -126,7 +127,15 @@ async function initProject(appName: string, options: CLIOptions) {
...
-      cloneAppAction(fullAppFolder, !!normalizedOptions.typescript);
+      await cloneAppAction(fullAppFolder, !!normalizedOptions.typescript);
+    }
+
+    if (
+      !isInteractive &&
+      isContentfulTemplate(templateSource) &&
+      normalizedOptions.deliveryFunction
+    ) {
+      await cloneDeliveryFunction(fullAppFolder, !!normalizedOptions.typescript);
```
kdamball, line 138: parallelize here too, with the action script above

packages/contentful--create-contentful-app/src/utils/file.ts
```diff
@@ -0,0 +1,47 @@
...
+}: MergeJsonIntoFileOptions): Promise<void> {
+  const sourceJson = await getJsonData(source);
+  const destinationJson = await getJsonData(destination);
+
+  const mergedJson = mergeFn(destinationJson, sourceJson);
+
+  await writeFile(resolve(destination), JSON.stringify(mergedJson, null, '  '));
+}
+
+export function exists(path: string): Promise<boolean> {
```
kdamball, line 43: nit: fwiw usually it's recommended to just try writing to the file directly and catch the `FILE_NOT_FOUND` equivalent (https://nodejs.org/api/fs.html#fsexistspath-callback). I guess here we kinda need it for a cleaner error

packages/contentful--create-contentful-app/src/index.ts
BasKiers: Personally I feel that in this instance the gained performance does not weigh up against the increased code complexity of parallelising these two tasks. As the consumer of the library, since I'm only calling this function once in the entire lifecycle of my application, I'd rather wait a few more seconds and have the logs read nicely step by step.

kdamball (approved):

ghepting:
Hey @BasKiers! moving forward can you please give @contentful/team-developer-experience a chance to review things before merging? And some additional context/screenshots/examples to help us understand and possibly contribute to the overall DevExp of new features like this would be splendid. We don't want to be forced to configure ourselves as required core reviewers, but we do now own and are responsible for this and other repositories that were previously managed by @contentful/team-extensibility and it's important that we start to work towards better awareness, communication, and coordination on changes that are going to be proposed or take place.

PR #1641, opened 2023-08-17
feat(deliveryFunctions): cca build and upload [MONET-1336]

kdamball:


packages/contentful--app-scripts/lib/clean-up/clean-up-bundles.test.js
```diff
@@ -14,7 +14,7 @@ const bundlesFixture = [
   { sys: { id: 'test-8' } },
 ];
 
-describe.only('cleanUpBundles', () => {
```
ghepting, line 17: oof lol

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -125,8 +125,68 @@ function getActionsManifest() {
...
+
+    if (!Array.isArray(manifest.deliveryFunctions) || manifest.deliveryFunctions.length === 0) {
+      return;
+    }
+
+    console.log('');
+    console.log(`  ----------------------------
+  Delivery functions found in ${chalk.bold(DEFAULT_MANIFEST_PATH)}.
+  ----------------------------`);
+    console.log('');
```
ghepting, line 154: we should really put this formatting in a logging function, huh?

kdamball: Yep - I remember we have something like this already somewhere

kdamball:
@BasKiers now I have all your commits 🙈

packages/contentful--app-scripts/package.json
```diff
@@ -27,7 +27,7 @@
     "prettier": "prettier **/*.js --write --ignore-path .gitignore",
     "lint": "eslint ./lib",
     "lint:fix": "npm run lint -- --fix",
-    "test": "mocha ./lib/**/*.test.js ./lib/*.test.js ./utils/**/*.test.js",
+    "test": "mocha \"./{,!(node_modules)/**/}*.test.js\" --exit",
```
andipaetzold, line 30: Why this change?

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -125,8 +127,64 @@ function getActionsManifest() {
...
-      )}.`
+        DEFAULT_MANIFEST_PATH,
+      )}.`,
+    );
+    // eslint-disable-next-line no-process-exit
+    process.exit(1);
+  }
+}
+
+function getDeliveryFunctionsManifest() {
```
andipaetzold, line 138: Could we somehow merge `getActionsManifest` and `getDeliveryFunctionsManifest` methods? They are both parsing the same file. When reading this code first, I was confused why we now have two manifests.

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -125,8 +127,64 @@ function getActionsManifest() {
...
+            'Error:',
+          )} Invalid IP address ${hasInvalidNetwork} found in the allowNetworks array for delivery function "${
+            deliveryFn.name
+          }".`,
+        );
+        // eslint-disable-next-line no-process-exit
+        process.exit(1);
+      }
+
+      // EntryFile is not used but we want to strip it from delivery functions
```
andipaetzold, line 172: grammar
```suggestion
      // EntryFile is not use, so we want to strip it from delivery functions
```

packages/contentful--app-scripts/package.json
kdamball: previous one was weird because every time you add a new folder you'd need to remember to pass the folder in here. I opted to just run everything with the `.test.js` suffix and also exit the test runner as previously we weren't even running them in the first place 🤦🏾

andipaetzold: The trick (I think) for mocha is to use quotes. Running `mocha "path/**/file.js"` is different to `mocha path/**/file.js`.

This also does the job and is maybe easier to understand as it doesn't have negation: 
```
mocha \"./{lib,utils}/**/*.test.js\" --exit
```

kdamball: That's exactly the problem I was trying to avoid - forgetting add new folders here because I'm pretty sure it will happen again 😅 . I'd be in favour of moving `utils` folder into the `lib` but that's probably a discussion for another PR

andipaetzold: +1 for only having one source folder.
I guess I have the same discussion in another PR atm: https://github.com/contentful/create-contentful-app/pull/1634#discussion_r1302874023

packages/contentful--app-scripts/lib/utils.js
```diff
@@ -73,7 +79,7 @@ const selectFromList = async (list, message, cachedOptionEnvVar) => {
   }
 };
 
-function getActionsManifest() {
+function getEntityFromManifest(type) {
```
andipaetzold, line 82: Alternatively, we could always parse both and return an object with two props 🤷

andipaetzold (approved):

packages/contentful--app-scripts/lib/utils.js
kdamball: But this way the function is smaller 😛

PR #1642, opened 2023-08-17
chore: downgrade prettier to v2 due to incompatibility with lerna v6

BasKiers:
It seems that the recently released v3.x.x version of `prettier` is [not compatible with](https://github.com/lerna/lerna/issues/3765) the v6.x.x version of `lerna`. 
Due to the breaking changes in `lerna` v7 it seems easier to pin `prettier` on v2.x.x for now and [unblock the pipeline](https://app.circleci.com/pipelines/github/contentful/create-contentful-app/8469/workflows/496e2a08-8677-43e1-a70a-79828e29a1c3/jobs/32104).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,9 @@ updates:
       - dependency-name: "@types/node"
         versions:
           - ">=15.0.0"
+      - dependency-name: "prettier"
+        versions:
+          - ">=3.0.0"
     commit-message:
       prefix: fix
       prefix-development: chore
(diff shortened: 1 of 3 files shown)


kdamball:
LGTM but I'd split the PRs to make DX's life easier

ghepting:
It looks like there are a lot of commits unrelated to this PR's title. can you please take a look and make this easier to review as @kdamball kindly suggested?

BasKiers:
I see that I did not correctly rebase to master, the PR should only contain the last commit

mgoudy91 (approved):

PR #1708, opened 2023-10-02
chore: upgrade to node lts (v18) [EXT-4812]

jjolton-contentful:
BREAKING CHANGE
- deprecate v14 and v16 CI workflows & add v20
- specify v18 in `engines` in package.json

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,19 +9,14 @@ environment: &base-env
   NPM_CONFIG_LOGLEVEL: warn
 
 executors:
-  node-container-14:
-    docker:
-      - image: cimg/node:14.15.0
-        environment: *base-env
-
-  node-container-16:
-    docker:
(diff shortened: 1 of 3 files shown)


t-col (approved):

t-col:
@jjolton-contentful we might want to make this release a new major version. If you add `BREAKING CHANGE` in the squashed commit description the `semantic-release` package should pick that up and create a v2 of this package

thoughts @contentful/team-developer-experience ?

mgoudy91:
> @jjolton-contentful we might want to make this release a new major version. If you add `BREAKING CHANGE` in the squashed commit description the `semantic-release` package should pick that up and create a v2 of this package
> 
> thoughts @contentful/team-developer-experience ?

I wasn't sure but saw this https://stackoverflow.com/a/66289428. Given that, I think you're totally right @t-col

ghepting (approved):

PR #1743, opened 2023-10-23
feat: assume ts for deliveryFunction if no option provided [MONET-1564]

pciazynski:
Previously, in order to use `--deliveryFunction` or `--action` option, user needed to specify also `--javascript` or `--typescript` option. If no such option was provided then regular interactive session was started.
After this change `--typescript` option will be automatically assumed. Typescript was chosen because it is already the default template for other options.

### DIFF
--- a/packages/contentful--create-contentful-app/src/getTemplateSource.ts
+++ b/packages/contentful--create-contentful-app/src/getTemplateSource.ts
@@ -92,7 +92,7 @@ async function makeContentfulExampleSource(options: CLIOptions): Promise<string>
     return selectTemplate(ContentfulExample.Typescript);
   }
 
-  if (options.deliveryFunction) {
+  if (options.deliveryFunction || options.action) {
     return selectTemplate(ContentfulExample.Typescript);
   }
 
--- a/packages/contentful--create-contentful-app/src/getTemplateSource.ts
+++ b/packages/contentful--create-contentful-app/src/getTemplateSource.ts
@@ -92,6 +92,10 @@ async function makeContentfulExampleSource(options: CLIOptions): Promise<string>
     return selectTemplate(ContentfulExample.Typescript);
   }
 
+  if (options.deliveryFunction) {
+    return selectTemplate(ContentfulExample.Typescript);
+  }
+
   return await promptExampleSelection();
 }
 
--- a/packages/contentful--create-contentful-app/src/index.ts
+++ b/packages/contentful--create-contentful-app/src/index.ts
@@ -114,7 +114,8 @@ async function initProject(appName: string, options: CLIOptions) {
       !normalizedOptions.example &&
       !normalizedOptions.source &&
       !normalizedOptions.javascript &&
-      !normalizedOptions.typescript;
+      !normalizedOptions.typescript &&
+      !normalizedOptions.deliveryFunction;
 
     const templateSource = await getTemplateSource(options);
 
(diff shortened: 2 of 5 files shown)


packages/contentful--create-contentful-app/src/getTemplateSource.ts
```diff
@@ -92,6 +92,10 @@ async function makeContentfulExampleSource(options: CLIOptions): Promise<string>
     return selectTemplate(ContentfulExample.Typescript);
   }
 
+  if (options.deliveryFunction) {
+    return selectTemplate(ContentfulExample.Typescript);
+  }
```
ghepting, line 97: I'm not sure if I understand the business logic here. What does using delivery functions have to do with template selection (js vs ts)?

packages/contentful--create-contentful-app/src/index.ts
```diff
@@ -135,7 +136,7 @@ async function initProject(appName: string, options: CLIOptions) {
       isContentfulTemplate(templateSource) &&
       normalizedOptions.deliveryFunction
     ) {
-      await cloneDeliveryFunction(fullAppFolder, !!normalizedOptions.typescript);
+      await cloneDeliveryFunction(fullAppFolder, !normalizedOptions.javascript);
```
ghepting, line 139: this change confuses me 🙃

packages/contentful--create-contentful-app/src/getTemplateSource.ts
pciazynski: It is indeed a bit confusing - the problem is that delivery functions are created side by side near the code for the frontend app as well.
We want to give the possibility to omit a `--javascript` or a `--typescript` flag. So if a user passes only the `--delivery-function` flag then we assume it will be Typescript for both delivery function code and the frontend code. I chose Typescript because it is already written in the help that the typescript is a default option.

packages/contentful--create-contentful-app/src/index.ts
pciazynski: It's needed to make the Typescript a default option.
The function signature is like that: `cloneDeliveryFunction(destination: string, templateIsTypescript: boolean)`
If we omit both flags I want Typescript by default, hence the change. Do you suggest to write it in a different way?

packages/contentful--create-contentful-app/src/getTemplateSource.ts
```diff
@@ -92,6 +92,10 @@ async function makeContentfulExampleSource(options: CLIOptions): Promise<string>
     return selectTemplate(ContentfulExample.Typescript);
   }
 
+  if (options.deliveryFunction) {
```
BasKiers, line 95: It is not part of the ticket description but I guess it would make sense to also implement the same logic for the `--action` flag to avoid confusion when developers try to run `create-contentful-app --action`
e.g.
```suggestion
  if (options.deliveryFunction || options.action) {
```

BasKiers (approved):
LGTM 🎉

packages/contentful--create-contentful-app/src/index.ts
kdamball: but this would use javascript as default? Doesn't it make sense something like `normalizedOptions.javascript ?? true`

kdamball (commented):
Additionally, lets change the test generating delivery functions to use the delivery function flag only without `--typescript`

packages/contentful--create-contentful-app/src/index.ts
pciazynski: I think it would be good if I change the function from 
`export async function cloneDeliveryFunction(destination: string, templateIsTypescript: boolean)` 
to 
`export async function cloneDeliveryFunction(destination: string, templateIsJavascript: boolean)` 
to make it less confusing?

kdamball: Makes it even easier :)

pciazynski: Done ;)

kdamball (approved):

packages/contentful--create-contentful-app/src/index.ts
ghepting: lol ty for helping my peanut brain with this :)

PR #1844, opened 2024-01-30
feat: change cli commands

cemreyuksel:
This PR changes delivery functions commands to functions. 

[Ticket](https://contentful.atlassian.net/browse/MONET-1715)


packages/contentful--create-contentful-app/src/index.ts
```diff
@@ -197,8 +197,8 @@ async function initProject(appName: string, options: CLIOptions) {
     )
     .option('-a, --action', 'include a hosted app action in the ts or js template')
     .option(
-      '-d, --delivery-function',
-      'include a hosted delivery function template (EAP only, contact monet@contentful.com if interested)',
+      '-f, --function',
+      'include a function template (EAP only, contact monet@contentful.com if interested)',
```
ghepting, line 201: Thus is not making sense to me...

ghepting (changes requested):
I'm very confused. This PR is just hijacking the top level namespace of functions while introducing no support or specificity in functions context or "accepts". I don't even see where the manifest is properly parsed to support not HAA and delivery functions. Did I overlook it?

packages/contentful--create-contentful-app/src/index.ts
ghepting: Monet is the contact for delivery functions only and we already have HAAs with app event functions and entry validation coming next

cemreyuksel: this PR is just for renaming delivery functions, additional support for accepts will be implemented differently. First we want to finish renaming without breaking anything and then implement rest of the changes.

cemreyuksel: Why does it not make sense? I think there is nothing wrong with using `function` instead of `hosted-delivery-function`. Do you prefer keeping `hosted`?

ghepting: no sorry, i don't understand the contact monet@ when you aren't the primary team owning *all the other function capabilities* is what i mean. there's already hosted app actions, too and many more to come shortly, so this feels wrong to land this monet@contentful.com contact information change

ghepting:
```suggestion
      'include a function template',
```

cemreyuksel: oh sorry, you are right. We are still in EAP that's why I didn't change this. Sure, updated!

packages/contentful--app-scripts/src/upload/create-app-bundle.ts
```diff
@@ -25,7 +25,8 @@ export async function createAppBundleFromUpload(settings: UploadSettings, appUpl
       appUploadId,
       comment: comment && comment.length > 0 ? comment : undefined,
       actions,
-      deliveryFunctions,
+      // TO DO: update once https://github.com/contentful/contentful-management.js/pull/2154 is merged.
```
ghepting, line 28: - [ ] do the todo before merging

ghepting (approved):

PR #1848, opened 2024-01-30
feat (@contentful/app-scripts): allow API host overrides when creating app definitions

t-col:
allows EU users to specify the EU version of Contentful's API as the host


jjolton-contentful (approved):
🎯

PR #1869, opened 2024-02-12
feat: add accepts parameter [MONET-1826]

cemreyuksel:
This PR adds `accepts` and validates the parameter that will be used in  `contentful-app-manifest.json` file. 

[Ticket](https://contentful.atlassian.net/browse/MONET-1826)

Nit: not adding `accepts` as an empty array unless in the manifest, can change it if's necessary.


packages/contentful--app-scripts/src/utils.ts
```diff
@@ -108,11 +108,17 @@ export function getEntityFromManifest<Type extends 'actions' | 'functions'>(type
...
 
+    const fieldMappingEvent = "graphql.field.mapping";
+    const queryEvent =  "graphql.query";
+
     const items = (manifest[type] as FunctionAppAction[] | ContentfulFunction[]).map((item) => {
       const allowNetworks = Array.isArray(item.allowNetworks)
         ? item.allowNetworks.map(stripProtocol)
         : [];
 
+      const accepts = 'accepts' in item ? Array.isArray(item.accepts) ? item.accepts : undefined : undefined;
```
t-col, line 119:
```suggestion
      const accepts = 'accepts' in item && Array.isArray(item.accepts) ? item.accepts : undefined;
```
Array.isArray(item.accepts) will return false if item.accepts is undefined

packages/contentful--app-scripts/src/utils.ts
```diff
@@ -125,13 +131,25 @@ export function getEntityFromManifest<Type extends 'actions' | 'functions'>(type
...
+        process.exit(1);
+      }
 
       // EntryFile is not used but we do want to strip it
       // eslint-disable-next-line no-unused-vars
       const { entryFile: _, ...itemWithoutEntryFile } = item;
 
       return {
         ...itemWithoutEntryFile,
+        ...(accepts !== undefined ? { accepts } : {}),
```
t-col, line 152:
```suggestion
        ...(accepts && { accepts }),
```
this will only add `{ accepts }` to the return if `accepts` exists

packages/contentful--app-scripts/src/utils.ts
```diff
@@ -108,11 +108,17 @@ export function getEntityFromManifest<Type extends 'actions' | 'functions'>(type
...
+    const fieldMappingEvent = "graphql.field.mapping";
+    const queryEvent =  "graphql.query";
+
     const items = (manifest[type] as FunctionAppAction[] | ContentfulFunction[]).map((item) => {
       const allowNetworks = Array.isArray(item.allowNetworks)
         ? item.allowNetworks.map(stripProtocol)
         : [];
 
+      const accepts = 'accepts' in item ? Array.isArray(item.accepts) ? item.accepts : undefined : undefined;
+      const hasInvalidEvent = accepts?.find((event) => ![fieldMappingEvent, queryEvent].includes(event));
```
mrlaessig, line 120: Nit: I would move this close to the `if` statement. So its faster to get to the definition when you read the `if` statement. 

```suggestion
      const hasInvalidEvent = accepts?.some((event) => ![fieldMappingEvent, queryEvent].includes(event));
```

As you are not looking for the event itself.

mrlaessig (approved):

t-col (approved):

PR #1946, opened 2024-04-18
chore: add external-references to the list of functions [MONET-1986]

pciazynski:
Ticket: https://contentful.atlassian.net/browse/MONET-1986


packages/contentful--create-contentful-app/src/includeFunction.ts
```diff
@@ -12,10 +12,11 @@ const addBuildCommand = getAddBuildCommandFn({
 });
 
 const VALID_FUNCTION_TEMPLATES_DIRS = [
-  'templates',
   'appevent-filter',
   'appevent-handler',
   'appevent-transformation',
+  'external-references',
+  'templates',
```
BasKiers, line 19: Can't we just remove this?

pciazynski: Probably we can, I am not sure what was the idea behind it...

BasKiers: The idea behind is is that in the `apps` repo the example templates are in `examples/functions-templates`. A previous PR changed the `--functions` parameter to grab the template to be applied from `examples/functions-${cliArg}`. Because this was not on the radar when creating the `function-templates` folder the `--function external-references` will need to look in the `function-templates` folder. AFAIK the `templates` option is only there as it satisfies the `function-${cliArg}` format even though `--function templates` does not mean anything to CLI users.

mgoudy91 (approved):

PR #1970, opened 2024-05-08
feat: adding cli feedback link + nudge

mgoudy91:
image

Adds feedback command `feedback` as well as a nudge on upload and activate commands to collect dev feedback


t-col:
so `create-contentful-app feedback` will just spit out the console.log with the typeform link right?

t-col (approved):

mgoudy91:
@t-col It will actually just open the user's default browser (if they have one) on that form

PR #2009, opened 2024-06-11
Feat: add a script to install the app

msieroslawska:
Based on the `open` script, I created a similar `install` script that opens the page where user can specify which space / environment they want the app to be installed to.


jjolton-contentful (approved):
💪

PR #2050, opened 2024-07-16
chore: update Node.js version requirement to >=20 in package.json

mgoudy91:
Recently found the app scripts bundle validator uses `readdirsync` with `recursive=true`

this is only available in node>=20

### DIFF
--- a/packages/contentful--app-scripts/package.json
+++ b/packages/contentful--app-scripts/package.json
@@ -10,7 +10,7 @@
     "directory": "packages/contentful--app-scripts"
   },
   "engines": {
-    "node": ">=14",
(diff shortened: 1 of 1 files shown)


jjolton-contentful:
Superseded by #3090, which bumps `@contentful/app-scripts` to `contentful-management` v12 and aligns `engines.node` to `>=20` across all packages in the repo (released as a breaking change).

Thanks for flagging the root cause here — the bundle validator's use of `fs.readdirSync(recursive: true)` (Node 20+ only) is exactly the kind of de-facto requirement that makes the `>=20` floor honest, and I've cited it in #3090's description. Closing this one out.

PR #2101, opened 2024-08-22
feat: changed readme to release new version []

mdmjg:
### Problem
- Packages relying on app-scripts are receiving a security alert about a vulnerable axios version

### Solution
- Release a new version of app-scripts so that packages using app-scripts can rely on non-vulnerable transitive dependencies

### Explanation
- There is a security vulnerability in axios that is fixed in version 1.7.4. 
- At some point, contentful-management.js was using another version. Contentful-management bumped the version, and app-scripts also bumped the version of contentful-management. 
- However, lerna is configured to ignore changes in `package.json` and thus did not release a new version. 
-  In this PR, we change the `README` so that a new version is released and the vulnerability is fixed.
- I chose to change the emoji to a woman emoji since I am a woman and thought it would be "very demure, very mindful" (this is a meme)


andipaetzold (approved):

Lelith (approved):

jjolton-contentful (approved):
🧑‍💻👩‍💻👨‍💻

PR #2111, opened 2024-08-27
feat: allow appaction.call events in the accepts field [EXT-5360]

jjolton-contentful:
A new Contentful Function event type (`appaction.call`) will soon be available. This change just ensures the app scripts tooling does not block uploads that contain app manifests with functions that accept these new events.


mgoudy91 (approved):
🤷

PR #2151, opened 2024-09-27
feat: strip protocol from host if provided during upload [EXT-00]

jjolton-contentful:
This PR addresses an issue where specifying a host that starts with http:// or https:// causes the upload process to break. 

To fix this, an inquirer filter function has been added to the host prompt in `buildAppUploadSettings`. This function strips out any leading http:// or https:// from the user's input, ensuring that only the base URL is used.

### DIFF
--- a/packages/contentful--app-scripts/src/upload/build-upload-settings.ts
+++ b/packages/contentful--app-scripts/src/upload/build-upload-settings.ts
@@ -37,7 +37,7 @@ export async function buildAppUploadSettings(options: UploadOptions): Promise<Up
       name: 'host',
       message: `Contentful CMA endpoint URL:`,
       default: DEFAULT_CONTENTFUL_API_HOST,
-      filter: (input: string) => input.replace(/^https?:\/\//, ''),
+      filter: hostProtocolFilter,
     });
   }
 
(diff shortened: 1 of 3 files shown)


packages/contentful--app-scripts/src/upload/build-upload-settings.ts
```diff
@@ -55,3 +56,7 @@ export async function buildAppUploadSettings(options: UploadOptions): Promise<Up
     ...appInfo,
   };
 }
+
+export function hostProtocolFilter (input: string) {
+  return input.replace(/^https?:\/\//, '');
```
t-col, line 61: does this catch `http://` as well?

jjolton-contentful: It does indeed but I probably should have added a test case for http too huh

```
Welcome to Node.js v20.17.0.
Type ".help" for more information.
> "http://api.contentful.com".replace(/^https?:\/\//, '')
'api.contentful.com'
> "https://api.contentful.com".replace(/^https?:\/\//, '')
'api.contentful.com'
```

jjolton-contentful: added in 483466d575b08240e144359fc3794d80a4cffcf4

t-col (approved):

PR #2153, opened 2024-09-30
feat: Support wildcard in Allow Networks MONET-2345

cemreyuksel:
[Ticket](https://contentful.atlassian.net/browse/MONET-2345)

We already support wildcard in allow networks Cloudflare. This PR allows users using this feature while creating an app and enables using wildcard domains in the `contentful-app-manifest.json` file. 

We use `tldts` package to parse the url and regexes to validate ip4 and ip6 formatted ip addresses.


cemreyuksel:
Closing this PR because we decided to use a regex to validate the networks.

PR #2176, opened 2024-10-28
feat: support setting navigationItem for page in create-app-definition [EXT-5884]

whitelisab:
When creating an app definition and having the page location selected, you can pass an optional `navigationItem` property which determines whether the page app should appear in the apps dropdown menu and what the name and path for that link should be. This is supported in the Contentful web app when creating and editing an app definition, but it is not supported in our `Create App Definition` script as part of the `@contentful/app-scripts` package. This PR adds this functionality to this script.

I have tested this locally and can confirm that the prompts appear correctly when the page location is selected and that the validations provide warnings correctly. The validations for the link name and path are using the same validations on the input fields in the web app.

Here is the experience in the web app:
Screenshot 2024-10-28 at 2 30 49 PM

Here is what the new prompts look like when using the `Create App Definition` script:
Screenshot 2024-10-28 at 2 24 44 PM

### DIFF
--- a/packages/contentful--app-scripts/src/create-app-definition/build-app-definition-settings.ts
+++ b/packages/contentful--app-scripts/src/create-app-definition/build-app-definition-settings.ts
@@ -87,8 +87,8 @@ NOTE: This will create an app definition in your Contentful organization.
       when(answers) {
         return answers.locations.includes('entry-field');
       },
-      validate(answers) {
-        if (answers.length < 1) {
+      validate(input) {
+        if (input.length < 1) {
           return 'You must choose at least one field type.';
         }
         return true;
@@ -107,13 +107,10 @@ NOTE: This will create an app definition in your Contentful organization.
       name: 'pageNavLinkName',
       message: 'Page location: Provide a name for the link in the main navigation:',
       when(answers) {
-        if (answers.locations.includes('page') && answers.pageNav) {
-          return answers.name;
-        }
-        return false;
+        return answers.locations.includes('page') && answers.pageNav;
       },
-      validate(answers) {
-        if (answers.length < 1 || answers.length > 40) {
+      validate(input) {
+        if (input.length < 1 || input.length > 40) {
           return 'Size must be at least 1 and at most 40';
         }
         return true;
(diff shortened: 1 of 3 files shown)


packages/contentful--app-scripts/src/create-app-definition/build-app-definition-settings.ts
```diff
@@ -87,14 +94,61 @@ NOTE: This will create an app definition in your Contentful organization.
...
+      when(answers) {
+        return answers.locations.includes('page');
+      },
+    },
+    {
+      name: 'pageNavLinkName',
+      message: 'Page location: Provide a name for the link in the main navigation:',
+      when(answers) {
+        if (answers.locations.includes('page') && answers.pageNav) {
+          return answers.name;
```
jjolton-contentful, line 111: I see this is working great as is but was thinking we might want:

```suggestion
          return !!answers.name;
```

just in case inquirer ever starts enforcing that the return value is really a boolean rather than just truthy

jjolton-contentful, line 111: does returning this value rather than a bool here have some special behavior? I couldn't find anything in the docs, was wondering if we might want to switch this to return a bool rather than something truthy in case inquirer gets more strict in the future

whitelisab: Yes, we should have this as a bool. I think this was leftover from some trial and error around trying to utilizing the result of the name prompt as a default here (I was unsuccessful 😞). So I will change this back to a boolean

packages/contentful--app-scripts/src/create-app-definition/build-app-definition-settings.ts
```diff
@@ -87,14 +94,61 @@ NOTE: This will create an app definition in your Contentful organization.
...
+        return answers.locations.includes('page') && answers.pageNav;
+      },
+      validate(answers) {
+        if (answers.length > 512) {
+          return 'Maximum 512 characters';
+        }
+        if (answers.includes(' ')) {
+          return 'Path cannot contain empty space';
+        }
+        if (!answers.startsWith('/')) {
```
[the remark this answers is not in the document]
whitelisab: ahh yes, that makes sense. Will adjust! 👍

jjolton-contentful (approved):

PR #2264, opened 2024-12-16
feat: build-functions app script [EXT-5978]

t-col:
adds a `build-functions` script to `contentful-app-scripts` to build Typescript, or non-standard Javascript, source code for Contentful Functions into an App Framework compatible bundle

### DIFF
--- a/packages/contentful--app-scripts/src/build-functions/build-functions.ts
+++ b/packages/contentful--app-scripts/src/build-functions/build-functions.ts
@@ -1,0 +1,127 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+import esbuild from 'esbuild';
+import path, { join, parse, resolve } from 'path';
+import { NodeModulesPolyfillPlugin } from '@esbuild-plugins/node-modules-polyfill';
+import { NodeGlobalsPolyfillPlugin } from '@esbuild-plugins/node-globals-polyfill';
+import { type BuildFunctionsOptions, type ContentfulFunction } from '../types';
+import { z } from 'zod';
+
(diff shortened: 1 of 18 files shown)


packages/contentful--app-scripts/src/build-functions/build-functions.ts
```diff
@@ -0,0 +1,127 @@
...
+      throw new Error(`Duplicate entryFile path: '${entryFile}'`);
+    }
+    if (uniqueValues.has(accepts)) {
+      throw new Error(`Duplicate accepts: '${accepts}'`);
+    }
+
+    uniqueValues.add(entryFile);
+    uniqueValues.add(path);
+    uniqueValues.add(id);
+    uniqueValues.add(accepts);
```
jjolton-contentful, line 56: I think duplicate `accepts` across functions shouldn't be a problem, was this maybe intending to ensure that accepts values themselves aren't duplicated within a single function?

I could envision a situation where it would be reasonable to have multiple functions that accept the same things, and a dev would switch between them as needed. For example, dev is adding a new function that is an improvement to an old function. Let's say both are `appevent.*`. Both do pretty much the same things, but dev wants to be able to quickly swap back if the new one doesn't work. They should be able to upload both and just switch between them using the app event subscription. As another example, users might want to build/upload a bunch of separate `appaction.call` functions, I think in its current form this validation would prevent that.

```suggestion

    uniqueValues.add(entryFile);
    uniqueValues.add(path);
    uniqueValues.add(id);

    const acceptsSet = new Set(accepts);
    if (acceptsSet.size !== accepts.length) {
      throw new Error(`Duplicate values found in 'accepts' for function with id '${id}'.`);
    }
```

jjolton-contentful (approved):

PR #2277, opened 2025-01-02
fix: improve error handling for invalid templates in CCA [EXT-4028]

whitelisab:
This PR addresses several improvements to error handling for create-contentful-app:

1. **Invalid template error:** Check to see if the template name provided with the `--example` option actually exists as a valid example. If not, then display a more helpful error message directing users where they can find the list.
2. **Invalid function template error:** When the user passes an invalid function template, clean up the error message so the formatting matches the invalid template error. Also, remove the cloned typescript example that is created before the function is created if there is an error.
3. **Error fetching templates:** Add error handling for fetching the example folder names

See below for before and after for these error messages.

| Before | After |
| --- | --- |
| **Invalid Template** | **Invalid Template** |
| Screenshot 2025-01-02 at 8 52 32 AM  |  Screenshot 2025-01-02 at 9 03 01 AM |
| **Invalid Function Template** | **Invalid Function Template** |
|  Screenshot 2025-01-02 at 8 56 13 AM |  Screenshot 2025-01-02 at 9 03 40 AM |
| **Error Fetching Templates** | **Error Fetching Templates** |
|  Screenshot 2025-01-02 at 8 48 24 AM  |  Screenshot 2025-01-02 at 9 01 51 AM  |

I also added new tests for all of these methods.

### DIFF
--- a/packages/contentful--create-contentful-app/src/getTemplateSource.ts
+++ b/packages/contentful--create-contentful-app/src/getTemplateSource.ts
@@ -1,22 +1,10 @@
-import fetch from 'node-fetch';
 import inquirer from 'inquirer';
 import chalk from 'chalk';
-import { CLIOptions, ContentfulExample } from './types';
-import { highlight, warn, wrapInBlanks } from './logger';
+import { CLIOptions, ContentfulExample, InvalidTemplateError } from './types';
+import { code, highlight, warn, wrapInBlanks } from './logger';
 import { EXAMPLES_PATH, IGNORED_EXAMPLE_FOLDERS } from './constants';
 import { isContentfulTemplate } from './utils';
-
-const CONTENTFUL_APPS_EXAMPLE_FOLDER =
-  'https://api.github.com/repos/contentful/apps/contents/examples';
-
-async function getGithubFolderNames() {
-  const response = await fetch(CONTENTFUL_APPS_EXAMPLE_FOLDER);
-  const contents = await response.json();
-
-  return contents
-    .filter((content: any) => content.type === 'dir' && !IGNORED_EXAMPLE_FOLDERS.includes(content.name))
-    .map((content: any) => content.name);
-}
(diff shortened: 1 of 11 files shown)


packages/contentful--create-contentful-app/src/getTemplateSource.ts
```diff
@@ -56,14 +44,20 @@ async function promptExampleSelection(): Promise<string> {
   } else {
     // get available templates from examples
     const availableTemplates = await getGithubFolderNames();
+    // filter out the ignored ones that are listed as templates instead of examples
```
whitelisab, line 47: This filtering is only relevant for the interactive example, so I moved it out of the `getGithubFolderNames` function

packages/contentful--create-contentful-app/src/includeFunction.ts
```diff
@@ -62,10 +66,9 @@ export async function cloneFunction(
     const packageJsonExists = await exists(packageJsonLocation);
 
     if (!packageJsonExists) {
-      console.error(
+      throw new Error(
```
BobHemphill76, line 69: I might need a bit of context here.  Where will this new error get caught?  And does that mean that line 87 should be modified as well?  For me, it's a bit weird that we have 3 different terminating styles in this  function.

BobHemphill76: Oh, nvm.   I see it now.  This throw is wrapped by the outer catch.  One thing to note after changing this flow is that it will now exit the process here, rather than log and continue processing.  Is that intentional?

whitelisab: Yes, I think we still want to exit the process here as it had been before if the function template is invalid. There was originally an exit on line 30 and I instead chose to throw the error that will be caught in this outer catch. If we don't exit, then the next items in the `initProject` function in `index.ts` will fail.

BobHemphill76 (approved):

PR #2325, opened 2025-02-07
fix: support host option for all commands [EXT-5971]

whitelisab:
This PR cleans up handling the `host` option throughout our `app-scripts` package. Here are the specific updates:

- `create-app-definition`: The copy displayed after app definition creation always displayed `app.contentful.com`. Now it will display the correct webapp host based on the host option.
- `upload`: Fixes a bug in interactive mode where the provided host option was not being properly passed through and also updates the copy after success to use the correct webapp host.
- `bundle-cleanup`: Fixes a bug in interactive mode where the provided host option was not being properly passed through.
- `activate`: Fixes a bug in interactive mode where the provided host option was not being properly passed through.
- `open-settings`: Host was not provided as an option, so this adds support for it.
- `install`: Fixes a bug where host value needed to be translated to web app host.

Now users should be able to use the host option for all of these commands to set the host to `api.eu.contentful.com` for EUDR customers.

### DIFF
--- a/packages/contentful--app-scripts/src/bin.ts
+++ b/packages/contentful--app-scripts/src/bin.ts
@@ -69,6 +69,7 @@ async function runCommand(command: Command, options?: any) {
     .command('open-settings')
     .description('Opens the app editor for a given AppDefinition')
     .option('--definition-id  [defId]', 'The id of your apps definition')
+    .option('--host [host]', 'Contentful domain to use')
     .action(async (options) => {
       await runCommand(open, options);
     });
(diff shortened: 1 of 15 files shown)


packages/contentful--app-scripts/src/bin.ts
```diff
@@ -69,6 +69,7 @@ async function runCommand(command: Command, options?: any) {
     .command('open-settings')
     .description('Opens the app editor for a given AppDefinition')
     .option('--definition-id  [defId]', 'The id of your apps definition')
+    .option('--host [host]', 'Contentful domain to use')
```
t-col, line 72:
```suggestion
    .option('--host [host]', 'Contentful API subdomain to use, e.g. "api.contentful.com"')
```

whitelisab: Thanks, good catch! I updated it for the all of the commands in this file

t-col (approved):

PR #2359, opened 2025-02-27
feat: pretty error message when app upload gets 403 [5872]

michaelpineirocontentful:
### Purpose
We want 403 messages to display a smaller, more informative message than a stack trace.


packages/contentful--app-scripts/src/upload/create-app-bundle.ts
```diff
@@ -27,7 +27,16 @@ export async function createAppBundleFromUpload(settings: UploadSettings, appUpl
       functions,
     });
   } catch (err: any) {
-    showCreationError('app upload', err.message);
+    try {
+      const message = JSON.parse(err.message);
+      if (message['status'] == 403) {
```
jjolton-contentful, line 32: wondering if it might be worth doing something like 
```suggestion
      if (message['status'] == 403 && message.details?.reasons) {
```

jjolton-contentful: to confirm the reasons are there before attempting to use them - just in case a weird 403 comes through that isn't shaped the right way

michaelpineirocontentful: Good catch, thank you!

jjolton-contentful (approved):

PR #2361, opened 2025-02-28
feat: examples come from new repository [EXT-6246]

michaelpineirocontentful:
### Purpose
We are moving our examples from ```apps``` to ```contentful-create-app-examples```.

### Approach
The current version is still pointed at v1, but can be altered easily via a constant.

1. Added ```-l, --legacy``` flag, which points to the previous version (in this case, still v1)
2. Added ```-n, --next``` flag, which points to the next version
3. Added ```-v, --version``` flag which allows you to point to any existing version. If the version doesn't exist, points to the current version.

I also made a change to the prompt logic. Instead of hardcoding templates in, we get the folders through GitHub (like we do for examples). I also made several semantic changes to avoid confusion caused by overlapping meanings to 'template.'

### Context
[Jira](https://contentful.atlassian.net/browse/EXT-6246)

### DIFF
--- a/packages/contentful--create-contentful-app/src/constants.ts
+++ b/packages/contentful--create-contentful-app/src/constants.ts
@@ -1,6 +1,6 @@
 export const CREATE_APP_DEFINITION_GUIDE_URL =
   'https://ctfl.io/app-tutorial#embed-your-app-in-the-contentful-web-app';
-export const EXAMPLES_REPO_URL = 'https://github.com/contentful/apps/tree/master/examples';
+export const REPO_URL = 'https://github.com/contentful/create-contentful-app-examples/tree/main';
 // These are the examples that are listed as templates instead of examples
 // OR should otherwise not be included in the list of examples displayed in the interactive mode
 export const IGNORED_EXAMPLE_FOLDERS = [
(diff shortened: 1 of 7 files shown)


packages/contentful--create-contentful-app/src/constants.ts
```diff
@@ -12,6 +12,25 @@ export const IGNORED_EXAMPLE_FOLDERS = [
...
 export const CONTENTFUL_APP_MANIFEST = 'contentful-app-manifest.json';
 export const IGNORED_CLONED_FILES = [CONTENTFUL_APP_MANIFEST, `package.json`];
+
+export const VERSION_1 = "1"
+export const VERSION_2 = "2"
+
+export const CURRENT_VERSION = VERSION_1 // what is currently being used
+export const LEGACY_VERSION = VERSION_1 // the last version that was used
+export const NEWEST_VERSION = VERSION_2 // the most recent version (not always current version)
+export const NEXT_VERSION = VERSION_2 // the next version that will be used (if current < newest)
```
david-shibley-contentful, line 36: beautiful 👨‍🍳

david-shibley-contentful (approved):
LGTM

packages/contentful--create-contentful-app/src/constants.ts
```diff
@@ -12,6 +12,26 @@ export const IGNORED_EXAMPLE_FOLDERS = [
   'hosted-app-action-templates',
   'function-templates',
 ] as const;
-export const EXAMPLES_PATH = 'contentful/apps/examples/';
+interface GithubPathFunction {
+  (version?: string): string;
+}
+
+export const examples_path: GithubPathFunction = function (version = CURRENT_VERSION) {
```
t-col, line 19: not normally a stickler for naming, but is there any reason not to use a standard camel case naming scheme here?

packages/contentful--create-contentful-app/src/getTemplateSource.ts
```diff
@@ -25,58 +25,58 @@ async function promptExampleSelection(): Promise<string> {
 
   // if the user chose to use a template, ask which language they prefer
   if (starter === 'template') {
+    const availableTemplates = await getGithubFolderNames(version, false);
+     // filter out the the function templates
+    const filteredTemplates = availableTemplates.filter(
+      (template) => !template.includes('function-')
+    );
```
t-col, line 32: worth adding a flag to `getGithubFolderNames` to DRY up this and line 47?

PR #2373, opened 2025-03-06
fix: no duplicate app manifests and better function name validation [EXT-6262]

michaelpineirocontentful:
### Purpose
Fix a small bug that adds a duplicate app-manifest to the ```functions``` folder when using ```generate-function```. Also, do early function name validation that will ensure the function can be uploaded.

### DIFF
--- a/packages/contentful--app-scripts/src/generate-function/build-generate-function-settings.ts
+++ b/packages/contentful--app-scripts/src/generate-function/build-generate-function-settings.ts
@@ -6,6 +6,7 @@ import { GenerateFunctionSettings, Language } from '../types';
 import ora from 'ora';
 import chalk from 'chalk';
 import { warn } from './logger';
+import { ValidationError } from './types';
 
 export async function buildGenerateFunctionSettingsInteractive() : Promise<GenerateFunctionSettings> {
   const baseSettings = await inquirer.prompt<GenerateFunctionSettings>([
@@ -14,9 +15,9 @@ export async function buildGenerateFunctionSettingsInteractive() : Promise<Gener
       message: `Function name (${path.basename(process.cwd())}):`,
     },
   ]);
-  if (BANNED_FUNCTION_NAMES.includes(baseSettings.name)) {
-    throw new Error(`Invalid function name: ${baseSettings.name}`);
-  }
+
+  validateFunctionName(baseSettings);
+
   const filteredSources = await getGithubFolderNames();
 
   const sourceSpecificSettings = await inquirer.prompt<GenerateFunctionSettings>([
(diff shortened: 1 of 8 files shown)


packages/contentful--app-scripts/src/generate-function/build-generate-function-settings.ts
```diff
@@ -14,8 +15,9 @@ export async function buildGenerateFunctionSettingsInteractive() : Promise<Gener
       message: `Function name (${path.basename(process.cwd())}):`,
     },
   ]);
-  if (BANNED_FUNCTION_NAMES.includes(baseSettings.name)) {
-    throw new Error(`Invalid function name: ${baseSettings.name}`);
+  if (BANNED_FUNCTION_NAMES.includes(baseSettings.name) || !(/^[a-z0-9]+$/i.test(baseSettings.name))) {
```
mgoudy91, line 18: This error won't make sense if you use a banned function name that IS alphanumberic

michaelpineirocontentful: Resolved

packages/contentful--app-scripts/src/generate-function/constants.ts
```diff
@@ -1,7 +1,7 @@
 export const EXAMPLES_PATH = 'contentful/apps/function-examples/';
 export const APP_MANIFEST = 'app-manifest.json';
 export const CONTENTFUL_APP_MANIFEST = 'contentful-app-manifest.json';
-export const IGNORED_CLONED_FILES = [APP_MANIFEST, `package.json`];
+export const IGNORED_CLONED_FILES = [APP_MANIFEST, CONTENTFUL_APP_MANIFEST, `package.json`];
```
michaelpineirocontentful, line 4: This is where the duplicate bug was

packages/contentful--app-scripts/src/generate-function/index.ts
```diff
@@ -3,14 +3,25 @@ import { buildGenerateFunctionSettingsInteractive, buildGenerateFunctionSettings
...
-
-  return create(generateFunctionSettings);
+  try {
+    const generateFunctionSettings = await buildGenerateFunctionSettingsInteractive();
+    return create(generateFunctionSettings);
+  } catch (e) {
+    if (e instanceof Error) {
+      console.log(e.message)
+    }
+  }
```
t-col, line 13:
```suggestion
  try {
    const generateFunctionSettings = await buildGenerateFunctionSettingsInteractive();
    return create(generateFunctionSettings);
  } catch (e) {
    if (e instanceof Error) {
      console.error(e.message)
    }
  }
```

packages/contentful--app-scripts/src/generate-function/build-generate-function-settings.ts
```diff
@@ -42,24 +42,38 @@ export async function buildGenerateFunctionSettingsInteractive() : Promise<Gener
...
     const optionKey = key as keyof GenerateFunctionSettings;
     const value = options[optionKey].toLowerCase().trim();
     
     if (optionKey === 'language') {
-      // Assert that the value is of type Language
       options[optionKey] = value as Language;
-    } else {
+    } else if (optionKey === 'example') {
       options[optionKey] = value;
+    } else { // don't want to lowercase function names
```
michaelpineirocontentful, line 75: This ensures that function names are not lowercased, as that may not be desired for the customer (eg. camel case)

packages/contentful--app-scripts/src/generate-function/index.ts
t-col: are we intentionally swallowing the error / not throwing here?

packages/contentful--app-scripts/src/generate-function/index.ts
```diff
@@ -3,14 +3,25 @@ import { buildGenerateFunctionSettingsInteractive, buildGenerateFunctionSettings
...
 
 const nonInteractive = async (options: GenerateFunctionSettings) => {
+  try {
     const generateFunctionSettings = await buildGenerateFunctionSettingsCLI(options);
     return create(generateFunctionSettings);
+  } catch (e) {
+    if (e instanceof Error) {
+      console.log(e.message)
+    }
+  }
```
t-col, line 24:
```suggestion
  } catch (e) {
    if (e instanceof Error) {
      console.error(e.message)
    }
  }
```

packages/contentful--app-scripts/src/generate-function/index.ts
```diff
@@ -3,14 +3,25 @@ import { buildGenerateFunctionSettingsInteractive, buildGenerateFunctionSettings
 import { create } from "./create-function";
 
 const interactive = async () => {
-  const generateFunctionSettings = await buildGenerateFunctionSettingsInteractive();
-
-  return create(generateFunctionSettings);
+  try {
+    const generateFunctionSettings = await buildGenerateFunctionSettingsInteractive();
+    return create(generateFunctionSettings);
+  } catch (e) {
```
michaelpineirocontentful, line 9: Previously I used ```process.exit(1)``` to exit the program when an error occured. However, this is very difficult to test. Instead, thrown errors are percolated to the top and logged here.

packages/contentful--app-scripts/src/generate-function/build-generate-function-settings.ts
```diff
@@ -42,24 +42,38 @@ export async function buildGenerateFunctionSettingsInteractive() : Promise<Gener
   return baseSettings
 }
 
-function validateArguments(options: GenerateFunctionSettings) {
+function validateFunctionName(baseSettings: GenerateFunctionSettings) {
```
michaelpineirocontentful, line 45: This function validates that the function name is alphanumeric and not banned

packages/contentful--app-scripts/src/generate-function/index.ts
michaelpineirocontentful: Yes - if an error is thrown by us (as Error) we don't usually want the stack trace. For example, if a user has a bad function name, they don't need a stack trace, just a notification that their name needs to be different

t-col (approved):

PR #2389, opened 2025-03-10
feat(app-scripts): windows compatibility

t-col:
Adds app-scripts compatibility for windows systems by removing hardcoded posix-standard paths in filesystem commands, in favor of dynamically resolved paths compatible with the host system.

Also enforces, through validation, that paths in the contentful-app-manifest.json file use posix-standard paths, regardless of host OS. These paths still get transformed to host-compatible paths when needed for filesystem commands.


packages/contentful--app-scripts/src/bin.ts
```diff
@@ -147,5 +147,6 @@ async function runCommand(command: Command, options?: any) {
   await program.parseAsync(process.argv);
 })().catch((e) => {
   console.error(e);
-  throw e
+  // eslint-disable-next-line no-process-exit
+  process.exit(1)
```
BobHemphill76, line 151: nice catch

packages/contentful--app-scripts/src/build-functions/__tests__/build-functions.test.ts
```diff
@@ -22,7 +21,7 @@ describe('resolveEsBuildConfig', () => {
     assert.strictEqual(Boolean(esbuildConfig), true);
     assertPartialObject(esbuildConfig, {
       entryPoints: {
-        'fixtures/function': `${path.resolve(__dirname, manifest.functions[0].entryFile)}`,
+        'fixtures/function': `./fixtures/function.ts`,
```
BobHemphill76, line 24: This was the esbuild issue you mentioned the other day?  They always assume posix pathing in their config?

BobHemphill76:
Whoa.  And the idea here is that once the manifest is off the client machine we are still always treating as POSIX down stream?

And I'm assuming we validated this on both a windows / mac machines (I see the CI is passing)

packages/contentful--app-scripts/src/build-functions/__tests__/build-functions.test.ts
t-col: yeah esbuild resolves modules rather than files, so we need posix/js module pathing here

PR #2415, opened 2025-03-26
fix: language-flags-with-function-templates [EXT-6291]

michaelpineirocontentful:
# Purpose
If someone uses `--javascript` or `--typescript`, they get a _template in that language._ Currently, if they also specify `--function <function-name>`, the function is _always_ typescript. This fixes that - if a user uses the template flag with `--function`, the function is in the language of the template flag.


whitelisab (approved):
Nice! ⭐

whitelisab (changes requested):
Actually, if you pass no flag, it creates a JS function with a ts app

whitelisab (approved):

PR #2419, opened 2025-03-28
feat: generate-function accepts keepPackageJson flag [EXT-6292]

michaelpineirocontentful:
### Purpose
The CCA uses generate-function to clone function templates. Current behavior does not preserve `package.json`. This change adds a flag (only to the CLI interface) to preserve `package.json`. 

Rather than merging with an existing `package.json` (normal behavior), this flag ensures that it is either created or overwrites the existing one. This option is intended to be used as a way to scaffold new FE-less apps.


packages/contentful--app-scripts/src/generate-function/clone.ts
```diff
@@ -78,17 +100,24 @@ export function resolvePaths(localPath: string) {
   return { localTmpPath, localFunctionsPath };
 }
 
-export async function cloneAndResolveManifests(cloneURL: string, localTmpPath: string, localPath: string, localFunctionsPath: string) {
+export async function cloneAndResolveManifests(cloneURL: string, localTmpPath: string, localPath: string, localFunctionsPath: string, keepPackageJson?: boolean) {
```
whitelisab, line 103: Maybe change this to have a default of `false` for `keepPackageJson`?

```suggestion
export async function cloneAndResolveManifests(cloneURL: string, localTmpPath: string, localPath: string, localFunctionsPath: string, keepPackageJson = false) {
```

packages/contentful--app-scripts/src/generate-function/clone.ts
```diff
@@ -78,17 +100,24 @@ export function resolvePaths(localPath: string) {
...
 
   // modify package.json build commands
-  await updatePackageJsonWithBuild(localPath, localTmpPath);
+  if (!keepPackageJson) {
+    await updatePackageJsonWithBuild(localPath, localTmpPath);
+  }
 
   // check if a tsconfig.json file exists already
-  const ignoredFiles = IGNORED_CLONED_FILES
+  const ignoredFiles = [...IGNORED_CLONED_FILES]
```
whitelisab, line 115: Does this need to be spread?

packages/contentful--app-scripts/src/generate-function/clone.ts
```diff
@@ -78,17 +100,24 @@ export function resolvePaths(localPath: string) {
...
   // merge the manifest from the template folder to the root folder
   await mergeAppManifest(localPath, localTmpPath);
 
   // modify package.json build commands
-  await updatePackageJsonWithBuild(localPath, localTmpPath);
+  if (!keepPackageJson) {
+    await updatePackageJsonWithBuild(localPath, localTmpPath);
+  }
 
   // check if a tsconfig.json file exists already
```
whitelisab, line 114: Looks like this comment belongs above line 121 instead

packages/contentful--app-scripts/src/generate-function/clone.ts
```diff
@@ -24,13 +24,15 @@ export async function cloneFunction(
     const { localTmpPath, localFunctionsPath } = resolvePaths(localPath);
 
     const cloneURL = getCloneURL(settings);
-    await cloneAndResolveManifests(cloneURL, localTmpPath, localPath, localFunctionsPath);
+    // Pass keepPackageJson if available in settings (from GenerateFunctionSettingsCLI)
+    const keepPackageJson = 'keepPackageJson' in settings ? (settings as any).keepPackageJson : undefined;
```
whitelisab, line 28: I think that we could avoid the typecasting as `any` by updating the `settings` argument type to reflect that it can be `GenerateFunctionSettings` or `GenerateFunctionSettingsCLI`

whitelisab (commented):
Added some comments for you to review. I think this approach makes sense; were you able to test that the existing behavior works as expected? 

Also, I assume because we only want to expose this `keep-package-json` to the CLI, that's why we don't need to add anything to the docs. However, I noticed that the `generate-functions` script is not in the `README` at all. Should we add it there? (can do it in a different PR if you want).

packages/contentful--app-scripts/src/generate-function/index.ts
```diff
@@ -16,7 +16,7 @@ const interactive = async () => {
   }
 };
 
-const nonInteractive = async (options: GenerateFunctionSettings) => {
+const nonInteractive = async (options: GenerateFunctionSettingsCLI) => {
   try {
     const generateFunctionSettings = await buildGenerateFunctionSettingsCLI(options);
     return create(generateFunctionSettings);
```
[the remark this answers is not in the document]
michaelpineirocontentful: This is fine because of [structural typing](https://www.typescriptlang.org/docs/handbook/type-compatibility.html) (i.e. `GenerateFunctionSettingsCLI` and `GenerateFunctionSettings` both share the same common subset of `name`, `example`, and `language`. I'm open to suggestions to change this if this obfuscates the meaning or makes the code hard to read.

packages/contentful--app-scripts/src/generate-function/clone.ts
michaelpineirocontentful: This is creating a copy of IGNORED_CLONED_FILES, because later in the code we are pushing to the list. Spreading is here to copy the list, rather than having a list-of-a-list. Another way to create a copy of this list could be with `const ignoredFiles = Array.from(IGNORED_CLONED_FILES);`. Let me know what you think!

packages/contentful--app-scripts/src/generate-function/index.ts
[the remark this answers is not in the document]
whitelisab: Nope, that works for me; thanks for clarifying!

packages/contentful--app-scripts/src/generate-function/clone.ts
whitelisab: That works and makes sense. I just saw it was different from where it was previously just using the constant itself and missed later in the code where you push to it. So let's keep this with what you have here!

whitelisab (approved):

PR #2434, opened 2025-04-04
fix: support bundles without frontend in upload and activate scripts [EXT-6304]

whitelisab:
## Overview

There was some default behavior in the `activate` script where the `src` property on a bundle was always being deleted and the `dialog` location was always being added. This makes sense if the app bundle you are uploading contains frontend files. But in order to support bundles with only Contentful functions, we needed to change this behavior. I added checks for whether the bundle has frontend files, and only delete the `src` property and add the `dialog` location if it does. Otherwise, if there is only functions code in the bundle, we should not perform those actions. This then allows the app developer to have both a `src` property (for local development or a third party URL) and a bundle with functions.

If the `activate` is being called from the `upload` script, we already have the bundle and can determine whether there are frontend files. If not, then we need to fetch this bundle.

## Testing

I added additional tests to verify this behavior. I also tested this locally with additional changes in our internal services and web app.

### DIFF
--- a/packages/contentful--app-scripts/src/activate/activate-bundle.test.ts
+++ b/packages/contentful--app-scripts/src/activate/activate-bundle.test.ts
@@ -1,7 +1,7 @@
 import { stub, match, SinonStub } from 'sinon';
 import assert from 'assert';
 import proxyquire from 'proxyquire';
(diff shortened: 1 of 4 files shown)


michaelpineirocontentful (approved):
⭐⭐⭐⭐⭐

PR #2518, opened 2025-06-05
fix: upgrade tiged dep rimraf - remove child dependency inflight [EXT-6447]

michaelpineirocontentful:
# Purpose
[Zendesk ticket](https://contentful.atlassian.net/browse/ZEND-6475)

## Testing
Tested this by merging to canary and using alpha release. `generate-function` (the script that uses tiged) is working as expected.


tylerwashington888 (approved):
LGTM

PR #2530, opened 2025-06-11
fix: remove child dependency inflight [EXT-6447]

michaelpineirocontentful:
The following dependency chain is causing vulnerabilities because inflight is deprecated and unstable:
tiged -> rimraf -> glob -> inflight.
Glob version > 9 deprecates inflight, rimraf version 5 uses glob v10.

Had a [similar change last week](https://github.com/contentful/create-contentful-app/pull/2518), as the issue was occurring in app-scripts. However, since the app-scripts version comes bundled with all the other packages (which had tiged OR rimraf < v5 OR another package in the repo as a dependency) we needed to update all packages in the repo to override rimraf.

In order to allow lerna to release, I made small adjustments to a few comments.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,5 +47,9 @@
   },
   "dependencies": {
     "open": "^10.1.0"
+  },
+  "overrides": {
+    "rimraf": "^5.0.0",
+    "glob": "^10.0.0"
   }
 }
--- a/packages/contentful--app-scripts/src/utils.ts
+++ b/packages/contentful--app-scripts/src/utils.ts
@@ -163,7 +163,7 @@ export function getFunctionsFromManifest(): Omit<ContentfulFunction, 'entryFile'
         process.exit(1);
       }
 
-      // EntryFile is not used but we do want to strip it
+      // EntryFile is not used but we do want to strip it 
       // eslint-disable-next-line no-unused-vars
       const { entryFile: _, ...itemWithoutEntryFile } = item;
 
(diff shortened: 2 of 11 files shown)


packages/contentful--app-scripts/src/utils.ts
```diff
@@ -163,7 +163,7 @@ export function getFunctionsFromManifest(): Omit<ContentfulFunction, 'entryFile'
         process.exit(1);
       }
 
-      // EntryFile is not used but we do want to strip it
+      // EntryFile is not used but we do want to strip it
```
BobHemphill76, line 166: nit: what's up with this formatting?

michaelpineirocontentful: Only added a space here so that lerna would detect code changes and allow a new release to go through

package.json
```diff
@@ -47,5 +47,9 @@
   },
   "dependencies": {
     "open": "^10.1.0"
+  },
+  "overrides": {
```
BobHemphill76, line 51: I guess this is ok, but I don't see tiged in the deps here?  Is the assumption tiged is transitive on one our dependencies?  If so, does the offending version of our dep have a sec vuln patch available?

packages/contentful--react-apps-toolkit/package.json
```diff
@@ -52,5 +52,9 @@
   "dependencies": {
     "contentful-management": ">=7.30.0"
   },
+  "overrides": {
```
BobHemphill76, line 55: same here, do we know the dep here that is offending?

t-col (approved):

PR #2631, opened 2025-08-05
feat(pnpm): support pnpm as package manager

theimbender:
This PR adds support for the `--pnpm` CLI option, and [pnpm](https://pnpm.io/) support in general, for the `create-contentful-app` tool. This PR **does not** convert the entire monorepo to use pnpm.

## Why is this needed?

For third party apps using pnpm as a package manager, trying to create Contentful apps with `create-contentful-app` via a custom [npm script](https://docs.npmjs.com/cli/v10/using-npm/scripts) (or any kind of script, really) can end up erroring out when the new app's dependencies are installed because of how different local versions of pnpm and npm are handled.

The commands given in console for installing dependencies, and for creating app definitions, are misleading in this case. There was also some ambiguity and unexpected behavior in the handling of when both `--yarn` and `--npm` flags were passed, and npm would be preferred even if it wasn't the current package manager. With this change, multiple package manager flags are ignored and the currently active package manager will be used; this is the same behavior as omitting a package manager flag at all.

This PR updates all areas that package managers are referenced, including the Segment.io analytics track call. Comprehensive test coverage is provided.

### DIFF
--- a/packages/contentful--create-contentful-app/README.md
+++ b/packages/contentful--create-contentful-app/README.md
@@ -30,6 +30,9 @@ npx create-contentful-app <app-name>
 # npm
 npm init contentful-app <app-name>
 
+# pnpm
+pnpm init contentful-app <app-name>
+
 # Yarn
 yarn create contentful-app <app-name>
 ```
@@ -38,11 +41,9 @@ yarn create contentful-app <app-name>
 
 ### Package Manager
 
-`--npm` or `--yarn`
+`--npm` or `--pnpm` or `--yarn`
 
-Use npm or Yarn to manage dependencies. If omitted, defaults to the manager used to run `create-contentful-app`.
-
-Both flags are mutually exclusive.
+Use npm, pnpm, or Yarn to manage dependencies. If omitted, or if more than one flag is passed, will default to the manager used to run `create-contentful-app`.
 
 ### Template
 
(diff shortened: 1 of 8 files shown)


michaelpineirocontentful (changes requested):
I made a few small tweaks, mostly to make `detectActivePackageManager()` more comprehensive. I have [opened a PR](https://github.com/theimbender/create-contentful-app/pull/1) against your branch. If you could review it and confirm it looks good, we can merge it in and then get this PR merged to main.

To ensure everything works as expected, I also [published an alpha release](https://github.com/contentful/create-contentful-app/releases/tag/create-contentful-app%402.0.16-alpha.0) that includes both of our commits. I tested it, and it worked successfully.

theimbender:
> I made a few small tweaks, mostly to make `detectActivePackageManager()` more comprehensive. I have [opened a PR](https://github.com/theimbender/create-contentful-app/pull/1) against your branch. If you could review it and confirm it looks good, we can merge it in and then get this PR merged to main.
> 
> To ensure everything works as expected, I also [published an alpha release](https://github.com/contentful/create-contentful-app/releases/tag/create-contentful-app%402.0.16-alpha.0) that includes both of our commits. I tested it, and it worked successfully.

Cool, I've tested on my local as well and it works as expected. I merged and just had one follow-up commit where I converted those requires to imports in the utils spec. Thank you for jumping on this!

michaelpineirocontentful (approved):

PR #3090, opened 2026-06-30
feat!: migrate app-scripts to contentful-management v12 (plain client, Node 20+)

jjolton-contentful:
## Summary

Migrates `@contentful/app-scripts` from `contentful-management` `^11.75.0` → `^12.8.0`, aligning it with the v12 already used by `@contentful/react-apps-toolkit` (and the app-sdk it depends on) and deduping `contentful-management` in `node_modules`.

CM v12 changed the default `createClient` overload to return the plain client (`PlainClientAPI`) instead of the nested (legacy) client. Rather than opting back into the deprecated legacy client, every `createClient` call site is moved onto the plain client — the legacy client is slated for removal in CM's next major, so this avoids a throwaway follow-up.

## 💥 Breaking change: drops Node 18 support

This is released as a **major** (`feat!`). `contentful-management` v12 requires **Node ≥20**, so `engines.node` moves `>=18` → `>=20` across all packages. Consumers on Node 18 (EOL) must upgrade.

Releasing this as a breaking change is consistent with how this repo has handled EOL Node drops before (e.g. [5007a4f](https://github.com/contentful/create-contentful-app/commit/5007a4f1d5ffae4c544e78d318e5cb436ea28219)). Note app-scripts' bundle validator already relies on `fs.readdirSync(recursive: true)` (Node 20+), so this also makes the declared minimum honest about what the code already requires (superseding the stale #2050).

Engines aligned to `>=20` in: `@contentful/app-scripts`, `@contentful/create-contentful-app`, the `create-contentful-app` CLI wrapper, and the workspace root.

The CI test matrix drops the `18.20` entry, adds Node 24 (Active LTS), and freshens the stale 20/22 pins: `['18.20', '20.18', '22.11']` → `['20.20', '22.23', '24.18']`. This stops testing a version the packages now declare unsupported and adds coverage for the current LTS. The required check (`CI Status`) rolls up the matrix via `needs: test`, so it isn't pinned to any per-version job.

### Call sites migrated to the plain client
| File | Before (nested) | After (plain) |
|---|---|---|
| `get-management-token.ts` | `client.getCurrentUser()` | `client.user.getCurrent()` |
| `organization-api.ts` | `client.getOrganizations()` / `getOrganization(id)` | `client.organization.getAll()` / `get({ organizationId })` |
| `definition-api.ts` | `org.getAppDefinitions()` / `getAppDefinition(id)` | `client.appDefinition.getMany(...)` / `get(...)` |
| `create-app-definition.ts` | `org.createAppDefinition(body)` | `client.appDefinition.create({ organizationId }, body)` |
| `upload/create-app-bundle.ts` | `org.getAppDefinition().createAppBundle()` | `client.appDefinition.get(...)` + `client.appBundle.create(...)` |
| `upload/create-app-upload.ts` | `org.createAppUpload(zip)` | `client.appUpload.create({ organizationId }, { file })` |

In `create-app-bundle.ts`, the org→appDefinition traversal the nested API required is replaced by a single `appDefinition.get` call kept as a verification step (preserves fail-early behavior on an invalid definition/org before the bundle is created).

## Test plan
- [x] `npm run build` (tsc) passes for `@contentful/app-scripts`
- [x] `npm test` — 177 passing; the 6 failures are pre-existing color-output assertion mismatches on `main`, unrelated to this change
- [x] `npm run lint` — 0 errors
- [x] Lockfile regenerated; `@contentful/app-scripts` resolves `contentful-management@12.8.0`
- [x] CI matrix drops Node `18.20`, adds `24.18` (Active LTS), freshens 20/22 pins → `['20.20', '22.23', '24.18']` 
 <div id='description'>
<a href="https://bito.ai#summarystart"></a>
<h3>Summary by Bito</h3>

<p>This PR migrates @contentful/app-scripts from contentful-management v11 to v12, upgrading the client API from the nested ClientAPI to the flat PlainClientAPI across all call sites. It also raises the minimum Node.js requirement from >=18 to >=20 across all workspace packages and the CI test matrix to align with contentful-management v12 requirements.</p>

<details>
<summary><i>Detailed Changes</i></summary>
<ul>

<li>Drops Node 18 support by raising engines.node from >=18 to >=20 across 4 packages (workspace root, @contentful/app-scripts, @contentful/create-contentful-app, and create-contentful-app CLI), reflecting contentful-management v12 requirements and existing fs.readdirSync(recursive:true) usage already present in the codebase.</li>

<li>Updates contentful-management dependency from ^11.75.0 to ^12.8.0 in @contentful/app-scripts package.json, aligning with @contentful/react-apps-toolkit and deduplicating the package in node_modules.</li>

<li>Migrates createClient calls in get-management-token.ts and get-app-info.ts to plain client by adding {type:'plain'} option and replacing client.getCurrentUser() with client.user.getCurrent().</li>

<li>Replaces organization-related nested API calls in organization-api.ts and definition-api.ts with flat plain client methods: getOrganization/getOrganizations become client.organization.get/getAll, and getAppDefinitions/getAppDefinition become client.appDefinition.getMany/get.</li>

<li>Updates upload module (create-app-bundle.ts, create-app-upload.ts) to use client.appDefinition.get for verification and client.appBundle.create/client.appUpload.create for resource creation instead of nested org methods like org.createAppUpload.</li>

</ul>
</details>

</div>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,7 +18,7 @@ jobs:
   test:
     strategy:
       matrix:
-        node-version: ['18.20', '20.18', '22.11']
+        node-version: ['20.18', '22.11']
         os: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,7 +18,7 @@ jobs:
   test:
     strategy:
       matrix:
-        node-version: ['20.18', '22.11']
+        node-version: ['20.20', '22.23', '24.18']
         os: [ubuntu-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
--- a/packages/contentful--app-scripts/package.json
+++ b/packages/contentful--app-scripts/package.json
@@ -55,7 +55,7 @@
     "bottleneck": "2.19.5",
     "chalk": "4.1.2",
     "commander": "12.1.0",
-    "contentful-management": "^12.3.1",
+    "contentful-management": "^12.8.0",
     "dotenv": "17.4.2",
     "esbuild": "^0.28.0",
     "ignore": "7.0.5",
--- a/packages/contentful--app-scripts/package.json
+++ b/packages/contentful--app-scripts/package.json
@@ -10,7 +10,7 @@
     "directory": "packages/contentful--app-scripts"
   },
   "engines": {
-    "node": ">=18",
+    "node": ">=20",
     "npm": ">=9"
   },
   "main": "lib/index.js",
@@ -55,7 +55,7 @@
     "bottleneck": "2.19.5",
     "chalk": "4.1.2",
     "commander": "12.1.0",
-    "contentful-management": "^11.75.0",
+    "contentful-management": "^12.3.1",
     "dotenv": "17.4.2",
     "esbuild": "^0.28.0",
     "ignore": "7.0.5",
--- a/package-lock.json
+++ b/package-lock.json
@@ -14269,7 +14269,7 @@
         "bottleneck": "2.19.5",
         "chalk": "4.1.2",
         "commander": "12.1.0",
-        "contentful-management": "^12.3.1",
+        "contentful-management": "^12.8.0",
         "dotenv": "17.4.2",
         "esbuild": "^0.28.0",
         "ignore": "7.0.5",
--- a/package-lock.json
+++ b/package-lock.json
@@ -6020,6 +6020,7 @@
       "version": "11.75.0",
       "resolved": "https://registry.npmjs.org/contentful-management/-/contentful-management-11.75.0.tgz",
       "integrity": "sha512-5j0NDwkdfj5x8qY7Ka6RsVMrjMSdu0clMbABQNqZA1wus9SPVETBQAzdJwHeOAhkLa6avzCcERlr4dN4lU/BOg==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "@contentful/rich-text-types": "^16.6.1",
@@ -14268,7 +14269,7 @@
         "bottleneck": "2.19.5",
         "chalk": "4.1.2",
         "commander": "12.1.0",
-        "contentful-management": "^11.75.0",
+        "contentful-management": "^12.3.1",
         "dotenv": "17.4.2",
         "esbuild": "^0.28.0",
         "ignore": "7.0.5",
@@ -14301,7 +14302,7 @@
         "ts-node": "10.9.2"
       },
       "engines": {
-        "node": ">=18",
+        "node": ">=20",
         "npm": ">=9"
       }
     },
@@ -14346,6 +14347,23 @@
       "license": "MIT",
       "engines": {
         "node": ">=18"
+      }
+    },
+    "packages/contentful--app-scripts/node_modules/contentful-management": {
+      "version": "12.8.0",
+      "resolved": "https://registry.npmjs.org/contentful-management/-/contentful-management-12.8.0.tgz",
+      "integrity": "sha512-lARtEZYn2MYf80hz9nCYiLbJWvg95TTnOWmCZouMZtgNS1273dN27Vaz7weHQ3oSQuk7LeYXaXdGFhFPKlpilw==",
+      "license": "MIT",
+      "dependencies": {
+        "@contentful/rich-text-types": "^16.6.1",
+        "axios": "^1.15.0",
+        "contentful-sdk-core": "^9.4.4",
+        "fast-copy": "^3.0.0",
+        "globals": "^15.15.0",
+        "process": "^0.11.10"
+      },
+      "engines": {
+        "node": ">=20"
       }
     },
     "packages/contentful--app-scripts/node_modules/define-lazy-prop": {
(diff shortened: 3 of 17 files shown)


jjolton-contentful:
Parking this as a draft. The deprecated legacy-client opt-in in here is only needed because the call sites still use the nested client — and migrating them to the plain client (the proper follow-up) requires the same CM v12 bump as its prerequisite.

Rather than ship the legacy stopgap and immediately revert it, we'll fold this bump into the plain-client migration so it lands once. Keeping the branch around as the verified starting point (build green, dedupe confirmed).

jjolton-contentful:
Tracked in AIS-83 (the proper plain-client migration this bump folds into).

packages/contentful--app-scripts/package.json
```diff
@@ -55,7 +55,7 @@
     "bottleneck": "2.19.5",
     "chalk": "4.1.2",
     "commander": "12.1.0",
-    "contentful-management": "^11.75.0",
+    "contentful-management": "^12.3.1",
```
whitelisab: The most recent version of this package is `12.8.0`, should we update this?

packages/contentful--app-scripts/package.json
```diff
@@ -10,7 +10,7 @@
     "directory": "packages/contentful--app-scripts"
   },
   "engines": {
-    "node": ">=18",
+    "node": ">=20",
```
whitelisab, line 13: When we've been dropping support for EOL versions of node in our other tools/SDKs, we have been releasing them as breaking changes ([see example](https://github.com/contentful/contentful-cli/pull/3193)). wdyt about that? Should we do that for this PR?

whitelisab (commented):
Two small questions for you to review! Thanks for updating all instances of using the legacy/waterfall client to the plain client!

packages/contentful--app-scripts/package.json
jjolton-contentful: Good call — done in 1cddb78d. Bumped the declared floor to `^12.8.0` so it matches what's already deduped in the lockfile. The caret was resolving to 12.8.0 anyway; this just makes the floor explicit and current.

packages/contentful--app-scripts/package.json
jjolton-contentful: Great question — I went back and forth on this one.

You're right that dropping an EOL Node version is conventionally a breaking change; that's exactly how we handled it in contentful-cli. The wrinkle here is the monorepo:

- nx is on `independent` versioning, **but** `@contentful/app-scripts` sits at the bottom of a dependency chain: `app-scripts` ← `@contentful/create-contentful-app` ← `create-contentful-app` (the CLI wrapper).
- A `BREAKING CHANGE` on app-scripts cascades version bumps up that chain, so we'd end up majoring the user-facing `create-contentful-app` CLI for what's really a Node-version policy bump on an internal build tool.
- That lockstep is visible in the current versions — those three are all on `3.0.1`, while `react-apps-toolkit` (which doesn't depend on app-scripts) sits at `2.0.1`.

So I kept it as `fix:` to avoid the cascade, and called out the Node 18 drop explicitly in the PR description so it isn't silent in the release notes.

That said, I'm not married to it. If you'd rather we eat the `create-contentful-app` CLI major bump for the sake of an honest semver signal (and consistency with how we treat EOL Node elsewhere), I'm happy to add the `BREAKING CHANGE:` footer. Curious which you see as the lesser evil here?

whitelisab: Thanks for the explanation! I think I'm still leaning towards making this a breaking change. It looks like the last time we dropped support for old node versions, we released it as a breaking change ([see PR](https://github.com/contentful/create-contentful-app/commit/5007a4f1d5ffae4c544e78d318e5cb436ea28219)). So then we would continue to be consistent in this repo. That's just my two cents on this!

jjolton-contentful: Sold — you're right, consistency with how we've handled EOL Node drops here wins. Thanks for digging up [5007a4f](https://github.com/contentful/create-contentful-app/commit/5007a4f1d5ffae4c544e78d318e5cb436ea28219).

Converted this to a breaking release:
- Retitled to `feat!:` and added a `BREAKING CHANGE:` footer (commit e3c70f7) so nx cuts a major.
- Aligned `engines.node` to `>=20` across all four packages + the root (they were drifting — the CLI wrapper was still on `>=14`).
- Bonus justification: app-scripts' bundle validator already uses `fs.readdirSync(recursive: true)` (Node 20+), so `>=20` is really just making the manifest honest. Closed the stale #2050 that originally spotted this.

Kept the scope lean on purpose — the pending runtime dep majors (inquirer/ora/commander/zod) don't clear any consumer-facing security advisories (only axios does, and that's a minor bump), so I'll handle those separately rather than bloat this major. Ready for another look when you have a sec!

whitelisab (dismissed):

.github/workflows/ci.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
   test:
     strategy:
       matrix:
-        node-version: ['18.20', '20.18', '22.11']
+        node-version: ['20.20', '22.23', '24.18']
```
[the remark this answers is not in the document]
jjolton-contentful: lots of users are still on 20 despite EOL and we still currently support it (https://github.com/contentful/contentful-management.js/blob/master/package.json#L23)

an unpatched Node CVE on a throwaway CI runner isn't a meaningful attack surface for create-contentful-app's consumers

whitelisab (approved):

PR #3145, opened 2026-08-05
feat(app-scripts): validate and preserve storage manifest declarations (PIC-1333)

ivo-toby:
## Summary
- Adds a strict Zod parser for the RFC-defined optional `storage` declaration in `contentful-app-manifest.json` (`src/manifest/storage.ts`), with a reusable Zod-issue formatter.
- `build-functions` now validates `storage.functions` against the manifest's declared Functions, wrapped in the existing command-facing error prefix.
- Both upload-setting builders (interactive and non-interactive/CI) preserve `storage` unchanged in `UploadSettings`.
- `validateBundle` re-checks `storage.functions` against upload settings' `functions`, covering the `upload-ci` route that can run without a preceding `build-functions`.
- Existing manifests without `storage` are completely unaffected at every integration point (dedicated no-storage tests throughout).

## Scope
Implements Tasks 1-5 of the plan in `docs/superpowers/plans/2026-08-05-pic-1333-storage-manifest-upload.md`. Task 6 (forwarding the declaration to Extensibility API via the App Bundle create call or a distinct CMA operation) is **intentionally not included** — it's blocked on a not-yet-published `contentful-management` client exposing the agreed operation (tracked against PIC-1339).

## ⚠️ Waiting on input
Two schema decisions in this PR are the implementer's tentative reading of an open RFC ambiguity, not yet confirmed  in the PIC-1333 Jira comment thread:
- `NAMESPACE_REGEX = /^[A-Za-z0-9_-]+$/` (allows `-`/`_`, keeping the RFC's `"content-ops"` example as-is) — isolated to one constant in `storage.ts`, a one-line change if he picks the strict-alphanumeric option instead.
- The column primary-key marker field is named `primaryKey`, and columns reject any `default` field (per the RFC's stated exclusion of custom SQL defaults) — same isolation, one schema.

Both are safe to merge as-is (self-contained, easily flipped) but should not be treated as final until Tyler responds.

## Test plan
- [x] `npm test --workspace @contentful/app-scripts -- --grep "storage|validateFunctions|validateBundle|buildAppUploadSettings|getUploadSettingsArgs"` — 36/36 passing
- [x] `npm run lint --workspace @contentful/app-scripts` — 0 errors
- [x] `npm run build --workspace @contentful/app-scripts` — clean
- [x] Full package suite: only 6 pre-existing failures in `generate-function/build-generate-function-settings.test.ts` (chalk/ANSI assertions), a file untouched by this branch — confirmed unrelated
- [x] Per-task implementer + reviewer pass for each of Tasks 1-5, plus a final whole-branch review — all clean, no Critical/Important findings 
 <div id='description'>
<a href="https://bito.ai#summarystart"></a>
<h3>Summary by Bito</h3>

<p>This PR adds strict Zod-based validation and propagation of the RFC-defined optional `storage` declaration in `contentful-app-manifest.json` across build-functions, upload, and validate-bundle operations. The implementation ensures that `storage.functions` references are validated against declared Functions at both build and upload stages, while preserving the storage declaration unchanged through upload settings for downstream API consumption.</p>

<details>
<summary><i>Detailed Changes</i></summary>
<ul>

<li>Introduces strict Zod schema validator in `storage.ts` for the `storage` declaration, enforcing column types (text, integer, real, boolean, json), table structure with primary key constraints, case-insensitive duplicate detection, reserved `_cf_` prefix validation, and server-side limits (64-char identifier max, 50 columns per table, 10 tables max), exported as `parseStorageDeclaration`, `assertStorageFunctionsKnown`, and `formatStorageValidationError`.</li>

<li>Refactors `validateFunctions()` in `build-functions.ts` to use the shared `assertStorageFunctionsKnown()` helper for cross-checking `storage.functions` against declared Function IDs, wrapping errors with the command-facing error prefix.</li>

<li>Updates both upload-setting builders (`buildAppUploadSettings` and `getUploadSettingsArgs`) to preserve the `storage` declaration unchanged in `UploadSettings`, passing it through to upload operations for downstream Extensibility API consumption.</li>

<li>Adds validation in `validateBundle()` that `storage.functions` references exist in upload settings' `functions`, covering the `upload-ci` route that runs without a preceding `build-functions` command.</li>

<li>Adds `getStorageFromManifest()` utility in `utils.ts` to read and validate the optional `storage` section from the manifest file, with error handling that prints Zod issue paths before exiting.</li>

</ul>
</details>

</div>

### DIFF
--- a/packages/contentful--app-scripts/src/build-functions/build-functions.ts
+++ b/packages/contentful--app-scripts/src/build-functions/build-functions.ts
@@ -6,6 +6,11 @@ import { NodeGlobalsPolyfillPlugin } from '@esbuild-plugins/node-globals-polyfil
 import { type BuildFunctionsOptions, type ContentfulFunction } from '../types';
 import { z } from 'zod';
 import { ID_REGEX, resolveManifestFile } from '../utils';
+import {
+  formatStorageValidationError,
+  parseStorageDeclaration,
+  type StorageDeclaration,
+} from '../manifest/storage';
 
 type ContentfulFunctionToBuild = Omit<ContentfulFunction, 'entryFile'> & { entryFile: string };
 
@@ -34,6 +39,19 @@ export const validateFunctions = (manifest: Record<string, any>) => {
     throw new Error(
       `Invalid Contentful Function manifest: ${JSON.stringify(validationResult.error.issues)}`
     );
+  }
+
+  let storage: StorageDeclaration | undefined;
+  try {
+    storage = manifest.storage === undefined ? undefined : parseStorageDeclaration(manifest.storage);
+    const functionIds = new Set(validationResult.data.functions.map((fn) => fn.id));
+    for (const functionId of storage?.functions ?? []) {
+      if (!functionIds.has(functionId)) {
+        throw new Error(`Storage declaration references unknown function id: '${functionId}'.`);
+      }
+    }
+  } catch (error) {
+    throw new Error(`Invalid Contentful Function manifest: ${formatStorageValidationError(error)}`);
   }
 
   const uniqueValues = new Set();
--- a/packages/contentful--app-scripts/src/build-functions/build-functions.ts
+++ b/packages/contentful--app-scripts/src/build-functions/build-functions.ts
@@ -7,6 +7,7 @@ import { type BuildFunctionsOptions, type ContentfulFunction } from '../types';
 import { z } from 'zod';
 import { ID_REGEX, resolveManifestFile } from '../utils';
 import {
+  assertStorageFunctionsKnown,
   formatStorageValidationError,
   parseStorageDeclaration,
   type StorageDeclaration,
@@ -45,11 +46,7 @@ export const validateFunctions = (manifest: Record<string, any>) => {
   try {
     storage = manifest.storage === undefined ? undefined : parseStorageDeclaration(manifest.storage);
     const functionIds = new Set(validationResult.data.functions.map((fn) => fn.id));
-    for (const functionId of storage?.functions ?? []) {
-      if (!functionIds.has(functionId)) {
-        throw new Error(`Storage declaration references unknown function id: '${functionId}'.`);
-      }
-    }
+    assertStorageFunctionsKnown(storage, functionIds);
   } catch (error) {
     throw new Error(`Invalid Contentful Function manifest: ${formatStorageValidationError(error)}`);
   }
--- a/packages/contentful--app-scripts/src/manifest/storage.ts
+++ b/packages/contentful--app-scripts/src/manifest/storage.ts
@@ -14,14 +14,26 @@ export const IDENTIFIER_REGEX = /^[A-Za-z_][A-Za-z0-9_]*$/;
  */
 const RESERVED_NAME_PREFIX = '_cf_';
 
+/**
+ * Identifier length and per-declaration count caps, mirroring
+ * `functions-api-client`'s `StorageDeclarationSchema` (PIC-1338) so a
+ * locally-valid declaration isn't later rejected by the authoritative
+ * server-side check.
+ */
+export const IDENTIFIER_MAX_LENGTH = 64;
+export const TABLE_MAX_COLUMNS = 50;
+export const TABLES_MAX_COUNT = 10;
+
 const storageColumnTypeSchema = z.enum(['text', 'integer', 'real', 'boolean', 'json']);
 
 // The RFC does not define a column `default` field (custom SQL defaults are
 // not expressible in the manifest), so `.strict()` rejects it along with any
-// other unrecognised key.
+// other unrecognised key. `default` will need to be revisited when schema
+// versioning/migrations land, since authoring migrations is explicitly out
+// of scope for this ticket (PIC-1333).
 const storageColumnSchema = z
   .object({
-    name: z.string().regex(IDENTIFIER_REGEX, 'Invalid column name'),
+    name: z.string().regex(IDENTIFIER_REGEX, 'Invalid column name').max(IDENTIFIER_MAX_LENGTH),
     type: storageColumnTypeSchema,
     nullable: z.boolean(),
     index: z.boolean().optional(),
@@ -31,17 +43,27 @@ const storageColumnSchema = z
 
 const storageTableSchema = z
   .object({
-    name: z.string().regex(IDENTIFIER_REGEX, 'Invalid table name'),
-    columns: z.array(storageColumnSchema).min(1, 'a table must declare at least one column'),
+    name: z.string().regex(IDENTIFIER_REGEX, 'Invalid table name').max(IDENTIFIER_MAX_LENGTH),
+    columns: z
+      .array(storageColumnSchema)
+      .min(1, 'a table must declare at least one column')
+      .max(TABLE_MAX_COLUMNS, 'a table may declare at most 50 columns'),
   })
   .strict();
 
 const storageDeclarationSchema = z
   .object({
     version: z.literal(1),
-    namespace: z.string().regex(IDENTIFIER_REGEX, 'Invalid namespace').optional(),
+    namespace: z
+      .string()
+      .regex(IDENTIFIER_REGEX, 'Invalid namespace')
+      .max(IDENTIFIER_MAX_LENGTH)
+      .optional(),
     functions: z.array(z.string()).min(1),
-    tables: z.array(storageTableSchema).min(1),
+    tables: z
+      .array(storageTableSchema)
+      .min(1)
+      .max(TABLES_MAX_COUNT, 'a storage declaration may declare at most 10 tables'),
   })
   .strict()
   .superRefine((storage, ctx) => {
@@ -130,6 +152,24 @@ export type StorageDeclaration = z.infer<typeof storageDeclarationSchema>;
 
 export const parseStorageDeclaration = (storage: unknown): StorageDeclaration =>
   storageDeclarationSchema.parse(storage);
+
+/**
+ * Shared cross-field check used by both the build-time Function manifest
+ * validation and the upload-time bundle validation: every Function id
+ * referenced in `storage.functions` must be present in `functionIds`. Throws
+ * on the first unknown id; callers are responsible for their own error
+ * wrapping/prefixing.
+ */
+export const assertStorageFunctionsKnown = (
+  storage: StorageDeclaration | undefined,
+  functionIds: Set<string>
+): void => {
+  for (const functionId of storage?.functions ?? []) {
+    if (!functionIds.has(functionId)) {
+      throw new Error(`Storage declaration references unknown function id: '${functionId}'.`);
+    }
+  }
+};
 
 export const formatStorageValidationError = (error: unknown): string => {
   if (error instanceof z.ZodError) {
--- a/packages/contentful--app-scripts/src/manifest/storage.ts
+++ b/packages/contentful--app-scripts/src/manifest/storage.ts
@@ -1,17 +1,10 @@
 import { z } from 'zod';
 
 /**
- * Namespace character set for the RFC "storage" declaration's `namespace`
- * field. Accepts hyphens and underscores in addition to alphanumerics,
- * distinct from the Function-ID charset (`ID_REGEX` in `../utils`).
- */
-export const NAMESPACE_REGEX = /^[A-Za-z0-9_-]+$/;
-
-/**
- * Identifier character set for RFC table and column names: a leading
- * letter or underscore, followed by letters, digits, or underscores.
- * Deliberately separate from `NAMESPACE_REGEX` so it can preserve the
- * underscore used by the RFC's own `brand_guidelines` example.
+ * Identifier character set for RFC table, column, and namespace names: a
+ * leading letter or underscore, followed by letters, digits, or underscores.
+ * Confirmed against the downstream Functions API's `StorageDeclarationSchema`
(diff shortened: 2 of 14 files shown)


packages/contentful--app-scripts/src/manifest/storage.ts
```diff
@@ -0,0 +1,147 @@
+import { z } from 'zod';
+
+/**
+ * Namespace character set for the RFC "storage" declaration's `namespace`
+ * field. Accepts hyphens and underscores in addition to alphanumerics,
+ * distinct from the Function-ID charset (`ID_REGEX` in `../utils`).
+ */
+export const NAMESPACE_REGEX = /^[A-Za-z0-9_-]+$/;
```
t-col: On the namespace charset question from the PR description: the RFC's contract section says namespace should match the same identifier restrictions as Function IDs (alphanumeric-only), but the RFC's own worked example uses `"content-ops"`, which has a hyphen. So this might be the RFC contradicting itself rather than an open ambiguity for you to resolve.

I don't imagine allowing hyphens/underscores causes a problem here — namespace only ever shows up inside an internal scope key, it doesn't need Function-ID-style uniqueness, and matching the worked example is probably what app authors will copy anyway. I'll go fix the RFC prose so this isn't ambiguous for future readers. Maybe we just keep `NAMESPACE_REGEX` as you have it?

packages/contentful--app-scripts/src/manifest/storage.ts
```diff
@@ -0,0 +1,147 @@
...
+ * with it (case-insensitive), regardless of `IDENTIFIER_REGEX`.
+ */
+const RESERVED_NAME_PREFIX = '_cf_';
+
+const storageColumnTypeSchema = z.enum(['text', 'integer', 'real', 'boolean', 'json']);
+
+// The RFC does not define a column `default` field (custom SQL defaults are
+// not expressible in the manifest), so `.strict()` rejects it along with any
+// other unrecognised key.
+const storageColumnSchema = z
```
t-col, line 35: Following up on the other open question from the description — the RFC's migration section lists adding a non-null column with a platform-supported constant default as valid additive schema evolution, and functions-api-app-storage's `ColumnDeclarationSchema` already carries a `default` field for exactly that (SQLite needs a default for a `NOT NULL` column added to a populated table).

Rejecting `default` here seems right for this ticket — `version` is pinned to the literal `1` and migration authoring is out of scope for PIC-1333. Might be worth a short comment or a linked follow-up so it doesn't get silently forgotten when schema versioning lands?

packages/contentful--app-scripts/src/manifest/storage.ts
```diff
@@ -0,0 +1,147 @@
...
+ */
+export const NAMESPACE_REGEX = /^[A-Za-z0-9_-]+$/;
+
+/**
+ * Identifier character set for RFC table and column names: a leading
+ * letter or underscore, followed by letters, digits, or underscores.
+ * Deliberately separate from `NAMESPACE_REGEX` so it can preserve the
+ * underscore used by the RFC's own `brand_guidelines` example.
+ */
+export const IDENTIFIER_REGEX = /^[A-Za-z_][A-Za-z0-9_]*$/;
```
t-col, line 9: One more, smaller: only the character set is checked for namespace/table/column names here, not length or count. functions-api-app-storage's `TableDeclarationSchema` caps identifiers at 64 chars and columns-per-table at 50, and extensibility-api's Function schema caps id/name at 64 too.

Not a bug today, but since PIC-1339's authoritative server-side check hasn't landed yet, a declaration that passes locally here could still get rejected once that check exists. Could be worth mirroring one of those existing limits?

t-col (commented):
Nice work! This is thorough — the Zod schema, the cross-field checks in build-functions and validate-bundle, and the no-storage-manifest coverage throughout all look solid.

My comments below are mostly around the two RFC ambiguities you flagged in the description (I think one of them is actually the RFC contradicting itself, not an open call for you to make), a couple of missing length/count caps that mirror limits already set elsewhere in functions-api-app-storage and extensibility-api, and a small duplication between build-functions and validate-bundle.

None of these feel blocking. If we choose not to address any of them now, it's probably worth a quick note in the PR description or a follow-up ticket so the context isn't lost once PIC-1339 picks this up.

packages/contentful--app-scripts/src/build-functions/build-functions.ts
```diff
@@ -36,6 +41,19 @@ export const validateFunctions = (manifest: Record<string, any>) => {
     );
   }
 
+  let storage: StorageDeclaration | undefined;
+  try {
+    storage = manifest.storage === undefined ? undefined : parseStorageDeclaration(manifest.storage);
+    const functionIds = new Set(validationResult.data.functions.map((fn) => fn.id));
+    for (const functionId of storage?.functions ?? []) {
```
t-col: The unknown-function-id check here is the same Set-based lookup and the same error message as in `validate-bundle.ts`. Maybe we pull it into a small shared helper in `manifest/storage.ts` (e.g. `assertStorageFunctionsKnown(storage, functionIds)`) so the two call sites can't drift apart later? Not blocking either way.

packages/contentful--app-scripts/src/utils.ts
```diff
@@ -186,6 +191,23 @@ export function getFunctionsFromManifest(): Omit<ContentfulFunction, 'entryFile'
   }
 }
 
+export function getStorageFromManifest(): StorageDeclaration | undefined {
```
t-col, line 194: Nit, feel free to ignore: `getFunctionsFromManifest` and `getStorageFromManifest` each independently read and `JSON.parse` the same manifest file, in both `build-upload-settings.ts` and `get-upload-settings-args.ts`. Harmless for a CLI tool — just flagging for completeness.

packages/contentful--app-scripts/src/manifest/storage.ts
ivo-toby: Thanks for looking! One wrinkle before I settle on this: I went and checked `functions-api` PR #2572 (PIC-1338), and `packages/functions-api-client/src/storage-declaration.ts`'s `StorageDeclarationSchema` validates `namespace` with the exact same `IDENTIFIER_PATTERN` it uses for table/column names (`/^[A-Za-z_][A-Za-z0-9_]*$/`) — no hyphens allowed there. So we've now got two signals from you pointing in opposite directions: this comment leans toward keeping the hyphen-permissive regex (matching the RFC's `content-ops` example), but your own shipped PIC-1338 code is strict and would reject it.

I already pushed a correction (33a8ddf3) matching the strict PIC-1338 pattern, on the theory that shipped code is more authoritative than an RFC's worked example. But given this comment, I'm not confident that's actually what you want here. Could you confirm which should govern — keep it strict to match PIC-1338, or make PIC-1338's namespace validation permissive too so they're consistent? Holding off on further changes to this regex until I hear back.

packages/contentful--app-scripts/src/manifest/storage.ts
ivo-toby: Good call — added a sentence to the comment above `storageColumnSchema` noting `default` will need revisiting once schema versioning/migrations land, since that's out of scope for PIC-1333. In 94347b37.

packages/contentful--app-scripts/src/manifest/storage.ts
ivo-toby: Mirrored functions-api-client's exact numbers (PIC-1338): IDENTIFIER_MAX_LENGTH=64 on namespace/table/column names, TABLE_MAX_COLUMNS=50, TABLES_MAX_COUNT=10. Added boundary tests for each (65-char name, 51 columns, 11 tables all rejected). In 94347b37.

packages/contentful--app-scripts/src/build-functions/build-functions.ts
ivo-toby: Extracted assertStorageFunctionsKnown(storage, functionIds) into manifest/storage.ts — both call sites use it now, and each keeps its own error-wrapping behavior (build-functions still re-prefixes with 'Invalid Contentful Function manifest:', validate-bundle still throws raw). In 94347b37.

packages/contentful--app-scripts/src/utils.ts
ivo-toby: Agreed, leaving this one as-is per your note — the two readers stay independent on purpose (documented in the code comment at both call sites).

packages/contentful--app-scripts/src/build-functions/build-functions.ts
```diff
@@ -36,6 +41,19 @@ export const validateFunctions = (manifest: Record<string, any>) => {
     );
   }
 
+  let storage: StorageDeclaration | undefined;
+  try {
+    storage = manifest.storage === undefined ? undefined : parseStorageDeclaration(manifest.storage);
+    const functionIds = new Set(validationResult.data.functions.map((fn) => fn.id));
```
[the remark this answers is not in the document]
ivo-toby: Already addressed — Tyler flagged the same duplication independently and it's fixed in 94347b37 (extracted assertStorageFunctionsKnown into manifest/storage.ts, used by both build-functions.ts and validate-bundle.ts). This comment predates that commit.

packages/contentful--app-scripts/src/upload/validate-bundle.ts
```diff
@@ -14,8 +15,11 @@ const fileContainsAbsolutePath = (fileContent: string) => {
...
 export const validateBundle = (
   path: string,
-  { functions }: Pick<UploadSettings, 'functions'>
+  { functions, storage }: Pick<UploadSettings, 'functions' | 'storage'>
 ) => {
+  const functionIds = new Set((functions ?? []).map(({ id }) => id));
+  assertStorageFunctionsKnown(storage, functionIds);
+
   const buildFolder = Path.join('.', path);
   const files = fs.readdirSync(buildFolder, { recursive: true, encoding: 'utf-8' });
```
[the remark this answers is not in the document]
ivo-toby: Not applying this one — buildFolder comes from --bundle-dir/the interactive prompt, supplied by the developer running this CLI locally or in their own CI. There's no trust boundary being crossed (same filesystem permissions as the process itself), so I don't think CWE-22 applies in this context.

packages/contentful--app-scripts/src/manifest/storage.ts
```diff
@@ -0,0 +1,180 @@
...
+    version: z.literal(1),
+    namespace: z
+      .string()
+      .regex(IDENTIFIER_REGEX, 'Invalid namespace')
+      .max(IDENTIFIER_MAX_LENGTH)
+      .optional(),
+    functions: z.array(z.string()).min(1),
+    tables: z
+      .array(storageTableSchema)
+      .min(1)
```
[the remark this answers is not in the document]
ivo-toby: This looks stale — .max(TABLES_MAX_COUNT, ...) is already applied to the tables array (storage.ts:63-66, same commit that introduced TABLES_MAX_COUNT). No change needed.

t-col (changes requested):

packages/contentful--app-scripts/src/manifest/storage.ts
ivo-toby: Kept `NAMESPACE_REGEX` as-is per your note — no code change needed here, just waiting on the RFC prose fix on your end.

packages/contentful--app-scripts/src/manifest/storage.ts
ivo-toby: Addressed in 94347b37: added a comment on `storageColumnSchema` noting `default` is deferred to schema versioning/migrations, out of scope for PIC-1333.

packages/contentful--app-scripts/src/manifest/storage.ts
ivo-toby: Addressed in 94347b37: added `IDENTIFIER_MAX_LENGTH` (64), `TABLE_MAX_COLUMNS` (50), `TABLES_MAX_COUNT` (10), mirroring functions-api-client's `StorageDeclarationSchema`.

packages/contentful--app-scripts/src/build-functions/build-functions.ts
ivo-toby: Extracted into `assertStorageFunctionsKnown` in 94347b37, now shared between build-functions.ts and validate-bundle.ts.

packages/contentful--app-scripts/src/utils.ts
ivo-toby: Leaving this as-is per your own note (harmless for a CLI tool) — deduplicating the manifest read across build-upload-settings.ts and get-upload-settings-args.ts means threading a shared parsed object through both, which is bigger than this ticket's scope. Can file a follow-up if it starts to matter.

ivo-toby:
All open review feedback on this PR is now addressed:

- Namespace regex, default-field rejection, identifier length/count limits, and the shared `assertStorageFunctionsKnown` helper — landed in 94347b37.
- Remaining bito-bot nits (manifest-access notation consistency, duplicate comment, misleadingly-named `RESERVED_NAME_PREFIX`) — landed in 8ff34dbb.
- The `getFunctionsFromManifest`/`getStorageFromManifest` duplicate-read nit is intentionally left as-is per the reviewer's own "harmless for a CLI tool" note — deduplicating it means threading a shared parsed manifest through both call sites, out of scope here.

All 5 open review threads are replied to and resolved. Ready for re-review.

PR #3159, opened 2026-09-09
fix: validate manifests with shared rules from node-apps-toolkit

michaelpineirocontentful:
## Summary

`app-scripts` implemented its own copies of the rules that decide whether an app manifest is valid — `allowNetworks` entries, App Action and Function ids, hosted code paths, action parameters — and those copies had drifted from the rules the API applies when the manifest is uploaded. The user-visible consequence: a wildcard domain with more than one subdomain label (`*.sub.sub.example.com`) was rejected locally even though it is a perfectly valid address, and some invalid manifests passed here only to fail on upload. This takes all of those rules from `@contentful/node-apps-toolkit`, which now owns them.

## Solution

- `isValidNetwork` now delegates to the shared rule. It keeps its name and signature — it is part of this package's published API — so this is not a breaking change for consumers.
- Replace the local `functionEvents` map with the shared event-topic check.
- Replace the local `parametersSchema` in `upsert-actions` with the shared parameter validator. As a side effect this fixes a real bug: the old zod schema marked `required` as non-optional, so it rejected parameters that `contentful-management` itself treats as valid.
- Derive the id-length error text from the shared bounds rather than hardcoding `1-64`, so the message cannot drift from the rule it describes.
- Close two build-time gaps in `build-functions`: a function `id` was checked for character set but never length, and `path` was not format-checked at all — both previously failed only after upload.
- Correct a test fixture that used `path: 'path/to/file.ts'`; `generate-function` writes `.js`, so the fixture was not a realistic manifest.

## Context

Same-shaped rules living in two codebases is what allowed the wildcard bug: one copy's wildcard branch matched a single subdomain label where the other allowed any number, so the two disagreed about the same address. Consolidating means a fix lands once and both sides pick it up on a version bump.

The shared rules are exported both as pattern strings and as predicates, so this package keeps validating with zod (`.refine(...)`) and does not have to adopt a different validation engine to share the definitions.

**Merge order:** this depends on `@contentful/node-apps-toolkit@^4.1.0`, which is not published yet — contentful/node-apps-toolkit#859 must merge and release first, and CI here will not resolve the dependency until it does. Verified locally against a packed tarball of that branch in the meantime.

## Test plan

- [x] `npm run build` clean against the shared package
- [x] Test suite: 191 passing, up from 181 on `main` (+10 new tests) with no new failures — the 6 pre-existing chalk color-code failures in `build-generate-function-settings.test.ts` are identical on untouched `main` and unrelated to this change
- [x] New tests cover the multi-label wildcard regression, long gTLDs, DNS label bounds, function id length at/over 64, path format, and a parameter with no `required` field
- [x] `npm run lint` — 0 errors
- [x] Confirmed `lib/utils.d.ts` still exports `isValidNetwork`, so the published type surface is unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/contentful--app-scripts/src/utils.ts
+++ b/packages/contentful--app-scripts/src/utils.ts
@@ -1,6 +1,7 @@
 import fs from 'node:fs';
 import chalk from 'chalk';
 import inquirer from 'inquirer';
+import { isValidNetworkAddress, isValidFunctionEventType } from '@contentful/node-apps-toolkit';
 import { cacheEnvVars } from './cache-credential';
 import { Definition } from './definition-api';
 import { Organization } from './organization-api';
@@ -10,18 +11,6 @@ import { resolve } from 'node:path';
 
 const DEFAULT_MANIFEST_PATH = resolve('.', 'contentful-app-manifest.json');
 
-const functionEvents = {
-  appActionCall: 'appaction.call',
-  appEventFilter: 'appevent.filter',
-  appEventHandler: 'appevent.handler',
-  appEventTransformation: 'appevent.transformation',
-  fieldMappingEvent: 'graphql.field.mapping',
-  resourceTypeMappingEvent: 'graphql.resourcetype.mapping',
-  queryEvent: 'graphql.query',
-  resourceLinksSearchEvent: 'resources.search',
-  resourceLinksLookupEvent: 'resources.lookup',
-};
-
 export const throwValidationException = (subject: string, message?: string, details?: string) => {
   console.log(`${chalk.red('Validation Error:')} Missing or invalid ${subject}.`);
   message && console.log(message);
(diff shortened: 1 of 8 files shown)


packages/contentful--app-scripts/src/utils.ts
```diff
@@ -1,6 +1,7 @@
 import fs from 'node:fs';
 import chalk from 'chalk';
 import inquirer from 'inquirer';
+import { isValidNetworkAddress, isValidFunctionEventType } from '@contentful/node-apps-toolkit';
```
jjolton-contentful: Since we only need validation, mind importing from `@contentful/node-apps-toolkit/validation` instead of root? Would keep us from dragging signing/JWT into the CLI dependency graph. Since your subpath export already supports it should just be a 3-file import path swap

michaelpineirocontentful: Done — switched all three imports to `@contentful/node-apps-toolkit/validation`. One catch: `eslint-plugin-node@11` doesn't resolve `package.json#exports` subpaths, so each import needs a scoped `eslint-disable-next-line node/no-missing-import` (matches the bare-disable style already used elsewhere in this file). Also had to add a `types` condition to every subpath in the toolkit's own `exports` map — TS's `node16` resolution doesn't fall back to a sibling `.d.ts` for subpaths the way plain Node resolution does, so the subpath resolved at runtime but not for the type-checker until that landed in node-apps-toolkit#859.

jjolton-contentful (approved):

